PR #1, opened 2022-09-23
start work on md parsing

SaraVieira:
adds simple markdown parsing that renders to public/index.html.

Does syntax hightlighting

tell me if trash

current tested markdown renders this
localhost_61730_


src/utils/syntax_highlight.rs
```diff
@@ -0,0 +1,13 @@
+use syntect::highlighting::ThemeSet;
+use syntect::html::highlighted_html_for_string;
+use syntect::parsing::SyntaxSet;
+
+pub fn syntax_highlight(lang: Option<&str>, code: &str) -> String {
+    let ps = SyntaxSet::load_defaults_newlines();
+    let ts = ThemeSet::load_defaults();
+    let language = if lang.is_none() { "rs" } else { "js" };
```
SaraVieira: Need to fix this, i forgot

src/utils/make_head.rs
```diff
@@ -0,0 +1,17 @@
+pub fn make_head() -> &'static str {
```
ashleygwilliams, line 1: Did you make these functions because you expect that there will be more logic later? If not I think you could probably make them constants.

src/utils/make_head.rs
```diff
@@ -0,0 +1,17 @@
...
+    let head = r#"
```
ashleygwilliams: it's idiomatic in rust to do returns without the `return` keyword- for this function I'd write:

```rust
fn my_func()-> &'static str {
   r#"stuff"
}
```

or alternatively, make it a constant:

```rust
static STUFF: &str = r#"stuff";
```

tests/main.rs
```diff
@@ -1,15 +1,15 @@
-use assert_cmd::Command;
-use oranda;
+// use assert_cmd::Command;
```
ashleygwilliams: unsure what the comments are for, my guess is you wanted them removed?

ashleygwilliams: oh wait i think i understand. before landing this i think would be cool to get these working. let me know if you want help!

src/utils/syntax_highlight.rs
```diff
@@ -0,0 +1,13 @@
...
+    let syntax = ps.find_syntax_by_extension(language).unwrap();
```
ashleygwilliams: we likely want to do some error handling instead of the unwraps, lemme know if you wanna pair on that

src/main.rs
```diff
@@ -1,4 +1,263 @@
-fn main() {
-    let msg = oranda::say_hello();
-    println!("{}", msg);
+use std::fs::File;
+use std::io::prelude::*;
+
+fn main() -> std::io::Result<()> {
+    let msg = oranda::create_html(
+        r#"
```
ashleygwilliams: this raw string is so long i would recommend making it a constant in a diff file or alternatively reading in an html file

src/lib.rs
```diff
@@ -1,3 +1,54 @@
...
+        fn build_pre_tag(&self, _attributes: &HashMap<String, String>) -> String {
+            String::from("")
+        }
+
+        fn build_code_tag(&self, _attributes: &HashMap<String, String>) -> String {
+            String::from("")
+        }
+    }
+
+    let adapter = MockAdapter {};
```
ashleygwilliams, line 57: you could probably do these 2 lines as a single line? i am definitely intrigued by this api looking at this- was this code you took from their example? does the adapter take options? (no judgement, just very curious)

ashleygwilliams (commented):
so first off - this is great.

literally all my comments are about rust idioms and cleaning things up/readability (which may be just be my personal aesthetics so please disagree eagerly). i think it would be helpful for you to run `cargo clippy` on this and checkout the feedback, i suspect it may catch at least part of what i've commented here. 

let me know if you have questions! i think the only thing i'd block on is getting `cargo clippy` to pass and then re-writing the commented out tests to test the functionality you've added here :)

src/utils/syntax_highlight.rs
ashleygwilliams: we can add this in a second, later, PR :)

src/utils/make_head.rs
SaraVieira: yes, this will later have a bunch of logic to make sure we have the right scripts and title, so that's why it's already moved.

Same thing for the footer one

src/utils/make_head.rs
SaraVieira: yeaahh, clippy warned me too. Fixed! :)

tests/main.rs
SaraVieira: Sure! Will try and ping you if I want to cry!

src/main.rs
SaraVieira: good point, completly forgot that. Moved to `test.md` in the root so its also easier to change

src/lib.rs
SaraVieira: yeah , its super weird. 

Here's the code: https://github.com/kivikakk/comrak/blob/main/src/parser/mod.rs#L475

I would love to clean it up because it also makes me add the `build_pre_tag` and `build_code_tag` that I do not need :/

src/utils/syntax_highlight.rs
SaraVieira: Sooo, I tried but now I have an error :(

https://user-images.githubusercontent.com/1051509/192157681-b2165cfe-5872-4d5c-bbf4-b96c08db7b5e.mp4

tests/main.rs
SaraVieira: I have been thinking about this and I don't actually think this makes sense to test since this package will not be a CLI but just an imported package.

It does make sense to test the lib and will do that :)

src/utils/syntax_highlight.rs
```diff
@@ -0,0 +1,35 @@
...
+    use super::*;
+    #[test]
+    fn parses_lists() {
+        /// syntax_highlight(Some("js"), "console.log(5)")?
+        assert_eq!(
+            syntax_highlight(Some("js"), "console.log(5)"),
+            Ok("".to_string())
+        );
+    }
+}
```
SaraVieira: trying to test this function and I get this error :(

PR #2, opened 2022-09-26
feat(sass) - Parse sass and create minified css file

SaraVieira:
This PR adds the ability to parse sass but it also changes the return type of the `create_html` function to be a tuple for the HTML and the css


Cargo.toml
```diff
@@ -10,3 +10,4 @@ assert_cmd="2"
 comrak = "0.14"
 syntect = "5.0"
 anyhow = "1.0"
+grass = "0.11.2"
```
ashleygwilliams, line 13: lol this crate name

src/css/_main.scss
```diff
@@ -0,0 +1,235 @@
+
```
ashleygwilliams: are the first two lines usually empty?

src/lib.rs
```diff
@@ -23,7 +24,7 @@ fn initialize_comrak_options() -> ComrakOptions {
     options
 }
 
-pub fn create_html(md: &str) -> String {
+pub fn create_html(md: &str) -> (String, String) {
```
ashleygwilliams: *personal opinion warning*
so i am not a huge fan of tuples in *almost* every case because i immediately forget what each thing in the tuple is, or what order things are in. especially because these are both string types- i would return a struct:

```rust
struct Site {
    html: String,
    css: String,
}

pub fn create(md: &str) -> Site {
...
```
*/warning*

ashleygwilliams (commented):
small comment on personal pref re: using a tuple vs a struct. extremely open to opposing views :)

src/css/_main.scss
SaraVieira: god no, let me clean that

Cargo.toml
SaraVieira: image

src/css/_main.scss
SaraVieira: fixed!

src/css/_tabs.scss
```diff
@@ -0,0 +1,78 @@
+.tabs {
```
ashleygwilliams, line 4: the other styles have a comment identifying them- this one didn't - wasn't sure if that was deliberate or not! (the comments seem nice, but i wouldn't trust my opinion heh)

src/css/_tacit.scss
```diff
@@ -0,0 +1,276 @@
+table {
```
ashleygwilliams, line 3: same as above q re: comment

src/lib.rs
SaraVieira: I do agree, it makes it way easier to tell by glacing at the code

Changed!

tests/lib.rs
```diff
@@ -1,32 +1,44 @@
 use oranda;
 
+#[test]
+fn returns_css() {
+    let site = oranda::create_html("# hello");
+    assert!(site
+        .css
+        .contains("--text-light:#fafafa;--text-800:#1f2937;"));
```
ashleygwilliams, line 8: not something to fix now, but all the `contains` here we'll probably want to move to some kind of fixture system or have them read directly from the files (e.g. i am imagining we change one of these colors and then the test breaks). i can file an issue if you think it's a good idea.

ashleygwilliams (commented):
i think this is looking good, 2 small comments:
- consistency re: commenting the css pieces
- eventually moving tests to fixtures and/or reading from the files themselves so we don't end up with spurious breaks because styles were updated

neither are blockers, i think this could be merged

src/lib.rs
ashleygwilliams: oooh, since it returns a "site" now- which is html and css- maybe the function name should be `create_site` ?

SaraVieira: renamed :)

src/css/_tabs.scss
SaraVieira: Added comments

src/css/_tacit.scss
SaraVieira: Done!

tests/lib.rs
SaraVieira: yes! Can you please make an issue?

sarahghp:
Hi! I'm a bit curious why we are bringing Sass in here. Is vanilla CSS not able to support something we expect? (It's really moving pretty quickly and capable of quite a lot these days.) When I look forward to this being configurable, I imagine something like many (too many) CSS variables and as flat a structure as possible, so that we avoid CSS complexity issues as well as we can. 

I would also think that using utility-style CSS would make it easier for users to intervene and change their generated code without running into problems with specificity (or being encouraged to do so). When we make our blog on the marketing site, we have to use `@apply` because of the way we target the markdown, but if we are generating the CSS & HTML, we definitely have other options open to us. Likewise, I wonder if it would be useful to use the generated CSS as an example for people who _do_ want to personalize. 

Finally, we probably want to be able to keep this in sync with our own design system for our properties, just for branding and product consistency reasons, and this looks like it is including values unlike ours (and somewhat arbitrary). 

Anyways, even if we don't end up making it the most extensible version to start off, which makes sense, I think it's worth considering some of these questions and slowly heading that way. In particular, I would recommend we use plain CSS, to reduce a dependency and to reduce the temptation to nest, which is likely to result harder-to-theme CSS in the future. 

(Also, I am sorry if this is a weird place to bring this up, but I can't think of a better one ...)

SaraVieira:
Hey!

Right now this is copying what was done in the node version, I used sass and would like to keep using it because of nesting mostly, nesting is love

I do believe I should make more variables so that's easy for people to change things but as of right now you can override our styles in the node version (https://axodotdev.github.io/oranda/#example-of-styles) and that will be converted into rust.

I am super okay in adding tailwind for making things easier in terms of using apply and maybe use the same spacing, as for the theme I am in the opionion the default theme should not be based on axo branding but we should and could make a theme based on it with our colors and fonts that a user could choose.

I will make an issue about CSS and ping you there and for now let's merge this as for now the main objective is feature parity

tests/lib.rs
SaraVieira: https://github.com/axodotdev/oranda-rs/issues/4

PR #9, opened 2022-09-29
feat - Add support for .oranda.config.json

### LINKED ISSUE #8: Allow to pass file parameter
We should able to allow passing a file parameter and that will be the file read.

It should default to `Readme.md || readme.md || README.md`

SaraVieira:
This PR adds support for getting the `.oranda.config.json` config and it for now reads:

- file
- dist folder
- name
- description
- thene

Need help with:

- I wanna connect twelf to clap but genuinly kinda stuck there as their examples are super simplistic

closes https://github.com/axodotdev/oranda-rs/issues/8


src/utils/options.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub file: String,
+    pub dist: String,
+    // pub name: String,
+    // pub logo: String,
+    // pub shareCard: String,
+    // pub homepage: String,
+    // pub noHeader: bool,
+    // pub theme: Theme,
+    // pub description: String,
+}
```
SaraVieira: Better way of doing this?

ashleygwilliams: Can you say more on what you are tryin to achieve- I suspect if the options are ... optional- than you should only need the top struct and not the "filled" one. Happy to jump on to pair if that'd help!

ashleygwilliams (commented):
happy to do a config pair!

src/utils/options.rs
SaraVieira: So I am trying to create default values for all of these and its just not letting me be happy :(

```rs
   let a = Options {
        file: options.file,
        dist: options.dist,
        ..Options::default()
    };
```

And I implemented a default fn for the Options:

```rs
impl Default for Options {
    fn default() -> Self {
        Options {
            file: Some("Readme".to_string()),
            dist: Some("public".to_string()),
        }
    }
}
```

It will still not send me the default values and if no value is passed to dist or file it's always empty :(

src/lib.rs
```diff
@@ -24,18 +25,32 @@ pub struct Report {
     // TODO: report useful paths/details for other tools
 }
 
-pub fn do_oranda() -> Result<Report> {
-    let mut file = File::open("test.md")?;
+pub fn do_oranda(options: Options) -> Result<Report> {
+    let defaults = Options::default();
+    let parsed_options = Options {
```
ashleygwilliams: *aesthetic opinion alert* i would potentially make this a function? just to break it up a bit */opinion*

src/lib.rs
```diff
@@ -24,18 +25,32 @@ pub struct Report {
...
+    };
+    let file = parsed_options.file.as_ref();
+    let mut file = File::open(file.unwrap())?;
     let mut data = String::new();
     file.read_to_string(&mut data)?;
-    let site = create_site(&data);
+    let site = create_site(&data, &parsed_options);
 
-    std::fs::create_dir_all("public")?;
+    let dist = &parsed_options.dist.unwrap_or_default();
```
ashleygwilliams: is this different from line 32?

src/lib.rs
```diff
@@ -24,18 +25,32 @@ pub struct Report {
...
 
-    let mut html_file = File::create("public/index.html")?;
+    std::fs::create_dir_all(&dist)?;
```
ashleygwilliams: *aesthetic opinion alert* i would potentially make the file operations a function? just to break it up a bit */opinion*

src/tests/mod.rs
```diff
@@ -1,36 +1,64 @@
-use crate as oranda;
+use crate::{
```
ashleygwilliams, line 1: i *think* you can say `use oranda;` and not the `crate` bit, style choice, i don't have a pref

ashleygwilliams (commented):
omg it helps to hit "finish your review" ashley.

src/lib.rs
SaraVieira: yeahhhh, I did that, that way its also everything in one file

src/tests/mod.rs
SaraVieira: I get an error :/

src/lib.rs
SaraVieira: Done!

src/lib.rs
SaraVieira: I moved this to a file so I think I lost what you meant

src/tests/mod.rs
ashleygwilliams: oh huh, i bet it has to do with a subtle diff in the way @Gankra and i do the lib/main split. i'll look into it, but yeah, as i said no worries

.gitignore
```diff
@@ -1,2 +1,5 @@
 /target
-public/
\ No newline at end of file
+public/
+
+# MAC
+.DS_Store
```
ashleygwilliams: this is an opinion i'm gonna recommend you ignore, lol, but i am sometimes pedantic about not putting local machine specific things in the project `.gitignore` and often will point people to use a global gitignore

src/lib.rs
```diff
@@ -1,13 +1,13 @@
-use std::{
-    collections::HashMap,
-    fs::File,
-    io::{Read, Write},
-};
+use std::{collections::HashMap, fs::File, io::Read};
```
ashleygwilliams: if you find it easier to read separated (like you had it previously), i often will just use multiple lines for my use statements, partially because it is much easier to remove them when you don't need them anymore, the diff is cleaner, etc. def an ✨ aesthetic ✨  choice that is oft debated in the rust community.

e.g.:

```rust
use std::collections::HashMap;
use std::fs::File;
```

ashleygwilliams (commented):
2 small things, definitely ignorable, mostly as an FYI

.gitignore
SaraVieira: Im gonna admit that I have never heard of a global gitignore

_googles_

SaraVieira: omg it worked, wow, wow

src/lib.rs
SaraVieira: I do agree it's easier to read, I guess its about finding a balance between the amount of lines and the readability

PR #17, opened 2022-09-30
re-orging options functionality

ashleygwilliams:


src/options/project/mod.rs
```diff
@@ -55,9 +57,10 @@ impl Rust {
 
 struct JavaScript {}
 impl JavaScript {
-    fn read(&self) -> Options {
-        let package_json = fs::read_to_string(PACKAGE_JSON).unwrap();
-        serde_json::from_str(&package_json).unwrap()
+    fn read(&self) -> Result<Options, OrandaError> {
+        let package_json = fs::read_to_string(PACKAGE_JSON)?;
+        let data: Options = serde_json::from_str(&package_json)?;
+        Ok(data)
```
ashleygwilliams: fun fact, if i don't separate it out like this, serde insists i write a deserializer for the error type

src/options/project/mod.rs
```diff
@@ -14,14 +15,14 @@ pub struct Options {
 }
 
 impl Options {
-    pub fn load() -> Option<Options> {
+    pub fn load() -> Result<Option<Options>, OrandaError> {
```
ashleygwilliams: return types like this make me want to rename Options to Config - thoughts @SaraVieira ?

SaraVieira: You don't like optional options? Ahah

SaraVieira: Super up for changing, any name ideas?

ashleygwilliams: was thinking "Config" just so we don't end up with an Option of Options that contain Option<T> ?

SaraVieira: I super approve this change

src/config/project.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    fn detect() -> Option<Type> {
+        if Rust::config().exists() {
+            Some(Type::Rust(Rust {}))
+        } else if JavaScript::config().exists() {
+            Some(Type::JavaScript(JavaScript {}))
+        } else {
+            None
+        }
```
Gankra: I assume long-term we'll want to detect if both are present and ask for clarification..?

Gankra (approved):

src/config/project.rs
ashleygwilliams: yeah 100%

ashleygwilliams:
ok, this has been open forever and is blocking further dev. i propose that this get merged and i have added the additional config functionality as issues.

ashleygwilliams:
issue was already filed https://github.com/axodotdev/oranda-rs/issues/11

SaraVieira:
I approve this idea

PR #21, opened 2022-12-12
fix description showing in body

SaraVieira:
Issue with quotes was making wrong html


src/site/html/head.rs
```diff
@@ -1,20 +1,21 @@
 use crate::config::{theme, Config};
 
 pub fn head(config: &Config) -> String {
+    println!("{:?}", &config.description);
```
ashleygwilliams, line 4: we probably want to take this out, but i'll leave it for now since we're not near shipping and i'll be doing a pass to make logging work

PR #22, opened 2022-12-13
Fix theme & dist_dir

SaraVieira:
- Fixes dist_dir by fixing a typo
- Fixes theme by allowing lowercase
- Adds axo theme that is present in the JS version


src/config/oranda.rs
```diff
@@ -6,7 +6,7 @@ use serde::Deserialize;
 use crate::config::theme::Theme;
 use crate::errors::*;
 
-static ORANDA_JSON: &str = "./oranda.json";
```
SaraVieira: Should I revert? I think adding the `.config` makes it clearer of what it is

src/config/mod.rs
```diff
@@ -70,7 +70,7 @@ impl Config {
                 // so return a merge of custom > project > default
                 return Ok(Config {
                     description: custom.description.unwrap_or(project.description),
-                    dist_dir: custom.dist_dir.unwrap_or(default.description),
+                    dist_dir: custom.dist_dir.unwrap_or(default.dist_dir),
```
ashleygwilliams, line 73: oh dang nice catch, we may want to add a test for these?

src/config/oranda.rs
ashleygwilliams: i'm less of a fan of this because it's not a pattern in Rust (i haven't seen a lot of tools that follow this pattern in general) i am open to suggestion tho.

src/config/theme.rs
```diff
@@ -1,14 +1,17 @@
 use serde::Deserialize;
 
 #[derive(Debug, Deserialize)]
+#[serde(rename_all = "lowercase")]
```
ashleygwilliams, line 4: i don't think we need this.. tell me more?

ashleygwilliams (commented):
some thoughts - can talk in 1:1

SaraVieira:
Merging this as conversations were fixed

PR #23, opened 2022-12-13
add possibility for additional css/sass files

### LINKED ISSUE #13: Allow to import other SCSS/Files
Allow the user to pass a `styles` options that points to a SCSS or CSS file, this file should be also parsed and minified in order to allow the user to add custom styles

SaraVieira:
Adds a new option called `additional_css` that you can use to pass a path to a local css or sass file and it will override our stuff

closes https://github.com/axodotdev/oranda/issues/13

related related https://github.com/axodotdev/oranda/issues/24

### DIFF
--- a/src/site/html/head.rs
+++ b/src/site/html/head.rs
@@ -5,7 +5,7 @@ pub fn head(config: &Config) -> String {
     format!(
         r#"
    <!DOCTYPE html>
-   <html lang="en">
+   <html lang="en" id="oranda">
    <head>
    
    <meta charset="utf-8" />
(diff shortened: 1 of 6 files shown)


src/site/html/head.rs
```diff
@@ -5,7 +5,7 @@ pub fn head(config: &Config) -> String {
     format!(
         r#"
    <!DOCTYPE html>
-   <html lang="en">
+   <html lang="en" id="oranda">
```
SaraVieira, line 8: moved so that people can target the body in their custom CSS

ashleygwilliams (approved):

Issue #24, opened 2022-12-14
Umbrella issue for feature parity


SaraVieira:
- [x] file → Your Readme.md name  
- [x] name → The project name that is in the title and the header  
- [x] logo → Link to your logo image, can be relative or absolute
- [x] description → The project description for meta tags     
- [x] shareCard → URL to social media preview image for meta tags (recommended size: 1200x628, URL cannot be relative) , impl https://github.com/axodotdev/oranda-js/blob/main/src/utils/head.js#L59
- [x] homepage →The project homepage for meta tags  (https://github.com/axodotdev/oranda/pull/30)
- [x] noHeader → Show no header and just the markdown content  impl https://github.com/axodotdev/oranda-js/blob/main/src/utils/header.js#L20
- [x] theme → Options are `light`, `dark`and `axo`     
- [x] syntaxHighlight → What syntax highlight theme to use in dark and light mode
- [x] favicon  → Favicon url or local path (impl https://github.com/axodotdev/oranda-js/blob/main/src/utils/head.js#L31)
- [x] dist → what folder to export your HTML   
- [x] styles → path to css/scss file (https://github.com/axodotdev/oranda/pull/23)                       
- [x] additionalFiles → Any other pages to create. It expects an array of paths of markdown files    
- [x] repository → Link to point the github corner  (https://github.com/axodotdev/oranda/pull/31)
- [x] pathPrefix → Host your oranda files at e.g. /my-oranda-project (impl https://github.com/axodotdev/oranda-js/blob/main/src/utils/header.js#L14)
- [ ] downloads → Links to download binaries or ways to install package, example below 
	Implementation:
	- 	https://github.com/axodotdev/oranda-js/blob/main/src/utils/createTabs.js
	- https://github.com/axodotdev/oranda-js/blob/main/src/commands/oranda.js#L95

Dropped

- [ ] ~~meta → Any extra meta tags you would like~~ 
- [ ]  ~~remoteScripts → Array of any remote scripts you want to include (eg: Google Analytics)~~
- [ ]  ~~remoteStyles → Array of any remote styles you want to include (eg: Google Fonts)~~

ashleygwilliams:
one box left! do we want to close this and open an issue for the final box (downloads)?

SaraVieira:
Yas! I'll close it as soon as the path prefix is merged :)

SaraVieira:
closed by https://github.com/axodotdev/oranda/issues/24

PR #25, opened 2022-12-14
add remote styles

SaraVieira:
- Adds possibility to pass an array of remote styles to be added to the CSS 

Need help with error handling and wondering if we should use the async version? I used this because it was the quickest since the async would need a bunch of changes in making everything be awaitable

example oranda.json for test:

```
{
  "theme": "axo",
  "remote_styles": ["https://unpkg.com/awsm.css@3.0.7/dist/awsm.min.css"]
}
```


ashleygwilliams (approved):

SaraVieira:
@ashleygwilliams Requested a review again because I moved the css function to its own file as it was getting quite chunky

src/site/css/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+
+pub fn build_css(config: &Config) -> Result<String> {
+    let css_options = grass::Options::default().style(OutputStyle::Compressed);
+    let mut css = grass::from_path("src/site/css/stylesheets/style.scss", &css_options)?;
+    if !config.additional_css.is_empty() {
+        if !Path::new(&config.additional_css).exists() {
+            return Err(OrandaError::FileNotFound {
+                filedesc: "Addicional CSS".to_string(),
+                path: config.additional_css.to_owned(),
+            });
```
SaraVieira: In the idea of error fast I also added this

ashleygwilliams: i think you have a spelling error here: "Addicional" -> "Additional"

src/config/project/javascript.rs
```diff
@@ -4,6 +4,7 @@ use std::path::{Path, PathBuf};
 #[cfg(test)]
 use assert_fs::fixture::{FileWriteStr, PathChild};
 
+#[allow(unused_imports)]
```
ashleygwilliams: can you say more about why we need this? in general i'd like to not include these

src/config/project/rust.rs
```diff
@@ -6,6 +6,7 @@ use serde::Deserialize;
 #[cfg(test)]
 use assert_fs::fixture::{FileWriteStr, PathChild};
 
+#[allow(unused_imports)]
```
ashleygwilliams: same here

src/site/css/mod.rs
```diff
@@ -0,0 +1,48 @@
+use crate::errors::*;
+use grass::OutputStyle;
+use std::fs::read_to_string;
+use std::path::Path;
+
+use crate::config::Config;
+
+pub fn build_css(config: &Config) -> Result<String> {
```
ashleygwilliams: since this is in a mod called "css" i would probably just call this "build" but this is a very tiny nitpick and is up to you to decide

src/site/css/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+    let css_options = grass::Options::default().style(OutputStyle::Compressed);
+    let mut css = grass::from_path("src/site/css/stylesheets/style.scss", &css_options)?;
+    if !config.additional_css.is_empty() {
```
ashleygwilliams: i would potentially break out some of these conditionals into smaller named functions to make it easier to test/read individual functionality, don't need to block on this tho, just sharing a stylistic pref

ashleygwilliams (commented):
a few small things- once the spelling error is fixed this can be merged, everything else was more discussing styles

src/site/css/mod.rs
SaraVieira: Fixed

src/config/project/javascript.rs
SaraVieira: oh, this is something I noticed when I downloaded the project, these imports would always give me warnings

src/site/css/mod.rs
SaraVieira: sounds good and done!

src/site/css/mod.rs
SaraVieira: broke both if statments into smaller functions

src/config/project/javascript.rs
ashleygwilliams: oh huh- i'll file an issue to review them, we probably want to fix the import instead of just allowing unused

ashleygwilliams (approved):
🚀

SaraVieira:
Removed the allow unused_imports so we remeber its there and will merge when ci passes

PR #26, opened 2022-12-16
feat(site): build site using typed html

### LINKED ISSUE #18: use typed-html instead of string concatenation

ashleygwilliams:
fixes #18 

~there's an outstanding type error that is very irritating, may need to release a new axohtml version to fix~ not actually a type error, thanks macros


ashleygwilliams:
i removed the homepage meta tag for now because how i handle the homepage needs to be reworked. figured i'd do it in a diff PR. will open an issue

SaraVieira (approved):

PR #27, opened 2022-12-16
add base for adding logo

### LINKED ISSUE #14: Allow user to pass a logo
Couple of things have to be done

- Read the logo option
- Check if image exists
- If exists add the header HTML code (JS equivalent: https://github.com/axodotdev/oranda/blob/main/src/utils/header.js)
- Copy image to the dist directory

SaraVieira:


src/config/mod.rs
```diff
@@ -117,6 +120,7 @@ impl Default for Config {
             theme: Theme::Light,
             remote_styles: vec![],
             additional_css: String::from(""),
+            logo: String::from(""),
```
ashleygwilliams: we probably want to make this `None`

src/site/html/head.rs
```diff
@@ -1,7 +1,13 @@
-use crate::config::{theme, Config};
+use crate::{
```
ashleygwilliams: i'd probably do this as 2 imports tbh

src/site/html/head.rs
```diff
@@ -1,7 +1,13 @@
...
+    config::{theme, Config},
+    errors::*,
+};
 
-pub fn head(config: &Config) -> String {
-    format!(
+use super::get_logo;
```
ashleygwilliams: if we can, i'd avoid using super imports- and ideally import it from it's mod

src/site/html/logo.rs
```diff
@@ -0,0 +1,52 @@
+use std::{
```
ashleygwilliams: i'd make this separate imports

src/site/html/logo.rs
```diff
@@ -0,0 +1,52 @@
...
+                    );
+
+                    let mut logo_file = File::create(&logo_path)?;
+                    logo_file.write_all(&img.bytes().unwrap())?;
+
+                    logo_url = &logo_path;
+                }
+            }
+        } else {
+            if !Path::new(&config.logo).exists() {
```
ashleygwilliams: this seems hard to read... maybe break it into a function?

src/site/html/logo.rs
```diff
@@ -0,0 +1,52 @@
...
+                return Err(OrandaError::FileNotFound {
+                    filedesc: "Logo".to_owned(),
+                    path: config.logo.to_owned(),
+                });
+            }
+
+            let new_path = Path::join(
+                Path::new(&config.dist_dir),
+                Path::new(&config.logo).file_name().unwrap(),
```
ashleygwilliams: can we replace the unwraps with ?s

ashleygwilliams (commented):
a few things since you pointed me in this direction ;)

src/config/mod.rs
SaraVieira: Done!

src/site/html/head.rs
SaraVieira: oh, didn't even notice, it was vscode

SaraVieira: fixed

src/site/html/logo.rs
SaraVieira: done!

src/site/html/logo.rs
SaraVieira: Can you leave this comment again where it was supposed to go? I changed things and have no idea what part you talking about?

src/site/html/logo.rs
SaraVieira: Big error


```
error[E0277]: the `?` operator can only be used on `Result`s, not `Option`s, in a function that returns `Result`
  --> src/site/html/logo.rs:30:89
   |
10 | / pub fn get_logo(config: &Config) -> Result<Option<&PathBuf>> {
11 | |     if !config.logo.is_none() {
12 | |         return Ok(None);
13 | |     }
...  |
30 | |                     Path::join(Path::new(&config.dist_dir), Path::new(&logo).file_name()?);
   | |                                                                                         ^ use `.ok_or(...)?` to provide an error compatible with `std::result::Result<std::option::Option<&PathBuf>, errors::OrandaError>`
...  |
54 | |     Ok(&logo_url)
55 | | }
   | |_- this function returns a `Result`
   |
   = help: the trait `FromResidual<std::option::Option<Infallible>>` is not implemented for `std::result::Result<std::option::Option<&PathBuf>, errors::OrandaError>`
   = help: the following other types implement trait `FromResidual<R>`:
             <std::result::Result<T, F> as FromResidual<Yeet<E>>>
             <std::result::Result<T, F> as FromResidual<std::result::Result<Infallible, E>>>


```

src/site/html/logo.rs
SaraVieira: Moved the remote logo fetching part, no idea if thats what you meant but it was chunky

src/site/html/logo.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+}
+
+pub fn get_logo(config: &Config) -> Result<Option<&PathBuf>> {
+    if !config.logo.is_none() {
+        return Ok(None);
+    }
+
+    let mut logo_url = None;
+    let logo = Option::expect(config.logo, "hmm, shouldn't really happen");
```
SaraVieira: Should this ever get here?

src/site/html.rs
```diff
@@ -20,9 +24,7 @@ pub fn build(config: &Config, content: String) -> String {
         </head>
         <body>
             <div class=classlist>
-                <div class="container">
```
SaraVieira: Pls don't delete, it breaks css 😢

ashleygwilliams: ok we should update the test then!

SaraVieira:
closed for https://github.com/axodotdev/oranda/pull/27

PR #29, opened 2022-12-17
[WIP] Assets

ashleygwilliams:
this is a PR off of #27 for building out some asset fetching abstractions for both local and remote logos, as well as CSS

to do:
- [x] copy local
- [ ] refactor local css to use it
- [x] copy remote
- [ ] refactor remote css to use it

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1222,6 +1222,7 @@ dependencies = [
  "tokio",
  "toml",
  "tower-http",
+ "url",
 ]
 
 [[package]]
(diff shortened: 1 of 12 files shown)


ashleygwilliams:
@SaraVieira when we let people pass relative URLs- they should be relative to what? most local project manifest file (e.g. Cargo.toml)? relative to the oranda.json file?

SaraVieira:
@ashleygwilliams I would say relative to the configuration file no? I would say relative to the oranda.json file

ashleygwilliams:
closing this in favor of making axo asset a separate lib, which we will integrate here iteratively

PR #30, opened 2022-12-19
Add homepage tag back

### LINKED ISSUE #28: add homepage meta tag back
removed in #26 because we need to rework what we do when no homepage is provided

SaraVieira:
Adds the homepage back

Requires https://github.com/axodotdev/axohtml/pull/8 to be merged

closes https://github.com/axodotdev/oranda/issues/28

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ license = "MPL-2.0"
 name = "oranda"
 
 [dependencies]
-axohtml = "0.1.0"
+axohtml = "0.2.0"
 axum = "0.5.17"
 clap = { version = "3.2.21", features = ["derive"] }
 comrak = "0.14"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ version = "0.0.0"
 edition = "2021"
 license = "MPL-2.0"
 
+
 [[bin]]
 name = "oranda"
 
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -5,11 +5,12 @@ version = "0.0.0"
 edition = "2021"
 license = "MPL-2.0"
 
+
```
ashleygwilliams: i think this is extraneous

src/site/html.rs
```diff
@@ -6,14 +6,21 @@ pub fn build(config: &Config, content: String) -> String {
     let theme = theme::css_class(&config.theme);
     let classlist: &str = &format!("body container {}", theme)[..];
     let description = &config.description;
+    let homepage = config.homepage.as_ref().map(|homepage| {
```
ashleygwilliams, line 10: does this resolve to an empty string if home page is `None`?

ashleygwilliams (commented):
i think this looks good- just remove the extra space. am curious the answer to my question! def wish we had that inline conditional

src/site/html.rs
SaraVieira: from what I could tell yes, no homepage and no html is rendered

Cargo.toml
SaraVieira: removed

ashleygwilliams (approved):
:shipit:

PR #33, opened 2022-12-20
fix(deps): address clapv4 deprecations

ashleygwilliams:


SaraVieira (approved):

SaraVieira:
We can also close the dependabot pr

PR #34, opened 2022-12-20
Improve syntax highlighting

SaraVieira:
- [x] Add padding arround code
- [x] add more themes
- [x] allow user to choose theme


SaraVieira:
Would recommend seeing PR like this: https://github.com/axodotdev/oranda/pull/34/files?file-filters%5B%5D=.md&file-filters%5B%5D=.rs&file-filters%5B%5D=.scss&show-viewed-files=true

It hides the PNG's and theme files

src/site/markdown/syntax_highlight.rs
```diff
@@ -1,34 +1,48 @@
...
+    }
+
+    if let Some(syntax_name) = syntax_name {
+        return Ok(syntax_name);
+    }
+
+    Err(OrandaError::Other(
+        "Please add the language to your code snippets".to_owned(),
+    ))
+}
```
SaraVieira: The reason for this function is that `find_syntax_by_extension` will work when your snippet uses `rs` but not when it uses `rust` as the language.

The other one works backwards so trash code it is

src/config/mod.rs
```diff
@@ -1,10 +1,12 @@
 mod oranda;
 mod project;
 pub mod theme;
-
+pub mod types;
```
ashleygwilliams: can we name this differently? maybe themes?

src/site/css/stylesheets/_main.scss
```diff
@@ -169,25 +169,16 @@ h6 {
...
   padding-left: 0;
   margin: 0;
-}
-
-.shiki {
   padding: 1em;
   border-radius: 4px;
 }
 
+
```
ashleygwilliams: extraneous space

src/site/markdown/mod.rs
```diff
@@ -10,8 +10,8 @@ use comrak::{self, ComrakOptions, ComrakPlugins};
 mod syntax_highlight;
 use syntax_highlight::syntax_highlight;
 
-pub struct MockAdapter {}
-impl SyntaxHighlighterAdapter for MockAdapter {
+pub struct Adapters {}
```
ashleygwilliams, line 12: oh nice

ashleygwilliams (changes requested):
so i think we should nest the syntax highlighting stuff in the markdown dir? since that is where it gets applied? curious what you think. otherwise, there's an extra space that should be removed. this looks good to me once that's handled and we move the theming stuff!

src/config/mod.rs
ashleygwilliams: on second though maybe we can nest this in `/markdown/syntax-highlight/` ?

src/site/markdown/syntax_highlight.rs
ashleygwilliams: this would be a good code comment here tbh

src/site/css/stylesheets/_main.scss
SaraVieira: fixed

src/config/mod.rs
SaraVieira: I moved a bunch of stuff, let me know what you think

src/site/markdown/syntax_highlight.rs
SaraVieira: done!

src/site/markdown/syntax_highlight/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use crate::config::Config;
+use crate::errors::*;
+use crate::site::markdown::syntax_highlight::syntax_themes::SyntaxThemes;
+use syntect::highlighting::ThemeSet;
+use syntect::html::highlighted_html_for_string;
+use syntect::parsing::{SyntaxReference, SyntaxSet};
+
+// The reason for this function is that find_syntax_by_extension will work when your snippet uses rs but not when it uses rust as the language.
+
```
ashleygwilliams: you'll want to put a `//` here to show it's a continuous comment

ashleygwilliams (approved):
very small request re the comment otherwise this is good to go

src/site/markdown/syntax_highlight/mod.rs
SaraVieira: Done!

PR #37, opened 2022-12-21
Add support for additional pages as markdown

SaraVieira:
Allows the user to pass any other markdown files that will also be created as HTML

When assets and logo are done we need to add the header that will add the menu links for these pages


ashleygwilliams (commented):
could you rebase this off of the assets pr? or should i jsut get that merged? it would be good for us to use that work here instead of duplicating it

SaraVieira:
I need both the assets and logo  :(

SaraVieira:
HTML issue created https://github.com/axodotdev/oranda/issues/46

ashleygwilliams (approved):
i am likely to change a lot of this once i merge #29 but for now this works

PR #38, opened 2022-12-21
Analytics

### LINKED ISSUE #36: Add support for analytics options
With not allowing for remote scripts we need to create a new option for allowing people to put their analytics scripts.

Pseudo code for this type:

```ts
type Analytics = {
	google: {
		tracking_id: String
	},
	plausible: {
		domain: String,
		script_url:  String  // Optional, defaults to https://plausible.io/js/script.js
	},
	fathom: {
		site: String,
	},
	unami: {
		website: String,
		script_url:  String,
	},
}
```

All of these should be optional

Ref: 
- https://chartio.com/learn/marketing-analytics/how-to-add-google-analytics-tracking-to-a-website/
- https://plausible.io/docs/plausible-script
- https://usefathom.com/docs/script/embed
- https://umami.is/docs/collect-data

SaraVieira:
Adds ability to configure analytics with the following options:

- Google
	- tracking_id
- Fathom
	- site
- Plausible
	- domain
	- script_url -> Optional, defaults to "https://plausible.io/js/script.js" 
- Unami
	- website
	- script_url 

closes https://github.com/axodotdev/oranda/issues/36


oranda.json
```diff
@@ -2,5 +2,10 @@
   "theme": "axo",
   "homepage": "https://oranda.axo.dev",
   "repository": "https://github.com/axodotdev/oranda",
+  "analytics": {
+    "plausible": {
+      "domain": "oranda.axo.dev"
```
ashleygwilliams: i strongly suspect this url is gonna be https://tools.axo.dev/oranda, but this is just used for testing ya?

src/config/analytics.rs
```diff
@@ -0,0 +1,80 @@
...
+    Plausible(PlausibleTracking),
+    Fathom(FathomTracking),
+    Unami(UnamiTracking),
+}
+
+const GOOGLE_SCRIPT_URL: &str = "https://www.googletagmanager.com/gtag/js";
+const PLAUSIBLE_SCRIPT_URL: &str = "https://plausible.io/js/script.js";
+const FATHOM_SCRIPT_URL: &str = "https://cdn.usefathom.com/script.js";
+
+pub fn get_google_script(config: &Config) -> Option<Box<script<String>>> {
```
ashleygwilliams: i might do this as an `impl GoogleTracking` since it's just related to that one and is using just that config

src/config/oranda.rs
```diff
@@ -8,6 +8,8 @@ use crate::errors::*;
 
 use crate::site::markdown::syntax_highlight::syntax_themes::SyntaxThemes;
 
+use super::analytics::Analytics;
```
ashleygwilliams: can we rewrite this to use something other than super? we use crate a lot and i figure it's good to be consistent

ashleygwilliams (changes requested):
small changes, feel free to push back or ask follow ups, mostly just prefs, nothing big

oranda.json
SaraVieira: yeah but might aswell do it right now

src/config/oranda.rs
SaraVieira: sure! Done, didn't catch it

src/config/analytics.rs
SaraVieira: Can you please give some follow up? I don't think I really understand the `impl` thing, some links would also be good so i could take a look tomorrow

SaraVieira: nvm I figured it out

let me know if it's okay

ashleygwilliams (commented):
if i am reading this correctly- if there isn't a google script- this would still call "get google script" and add it to the page? if we want to merge this we need to open an issue to fix that because not every page will use a google script.

it would also be helpful for us to get some tests for this! but we can open that as a separate issue

ashleygwilliams (approved):
good for now, based on pairing

Issue #48, opened 2022-12-23
remove all unwraps and replace with error handling


ashleygwilliams:

SaraVieira:
Current unwraps:

- [ ] https://github.com/axodotdev/oranda/blob/main/src/site/mod.rs#L23
- [ ] https://github.com/axodotdev/oranda/blob/main/src/site/mod.rs#L49
- [x] https://github.com/axodotdev/oranda/blob/main/src/site/mod.rs#L62
- [ ] https://github.com/axodotdev/oranda/blob/main/src/site/css/mod.rs#L39
- [ ] https://github.com/axodotdev/oranda/blob/main/src/site/markdown/syntax_highlight/mod.rs#L35

manyinsects:
Closing for now, we can take another look at this at some future point, but the current state of things looks pretty good.

PR #50, opened 2022-12-23
Remove unwrap

SaraVieira:
Removes an uneeded unwrap, fixed by matching and just doing nothing in the case there are no additional files

related https://github.com/axodotdev/oranda/issues/48

### DIFF
--- a/src/site/mod.rs
+++ b/src/site/mod.rs
@@ -59,8 +59,10 @@ impl Site {
         let site = Self::build(config, readme_path)?;
         let dist = &config.dist_dir;
         let mut files = vec![readme_path];
-        if config.additional_pages.is_some() {
-            files.extend(config.additional_pages.as_ref().unwrap())
+
+        match &config.additional_pages {
+            Some(pages) => files.extend(pages),
+            None => {}
         }
 
         for file in files {


src/site/mod.rs
```diff
@@ -58,8 +58,10 @@ impl Site {
         let site = Self::build(config, readme_path)?;
         let dist = &config.dist_dir;
         let mut files = vec![readme_path];
-        if config.additional_pages.is_some() {
-            files.extend(config.additional_pages.as_ref().unwrap())
+
+        match &config.additional_pages {
```
ashleygwilliams, line 63: i think an if let is probably a better fit here if we dont intend to handle the None case

ashleygwilliams (changes requested):

src/site/mod.rs
```diff
@@ -58,8 +58,9 @@ impl Site {
         let site = Self::build(config, readme_path)?;
         let dist = &config.dist_dir;
         let mut files = vec![readme_path];
-        if config.additional_pages.is_some() {
-            files.extend(config.additional_pages.as_ref().unwrap())
+
+        if let Some(pages) = config.additional_pages {
+            files.extend(pages);
```
SaraVieira: Changed it here but I get a weird error: 

CleanShot 2023-01-06 at 18 45 21@2x


cc @ashleygwilliams

ashleygwilliams: oh huh mmm - mind if i pull this down and poke at it?

SaraVieira: Of course not :)

src/site/mod.rs
```diff
@@ -55,17 +55,19 @@ impl Site {
     }
 
     pub fn write(config: &Config) -> Result<()> {
-        let readme_path = &config.readme_path;
-        let site = Self::build(config, readme_path)?;
+        let readme_path = config.readme_path.clone();
```
ashleygwilliams: smol crime, it's fine

src/site/mod.rs
```diff
@@ -55,17 +55,19 @@ impl Site {
...
+        let site = Self::build(config, &readme_path)?;
         let dist = &config.dist_dir;
-        let mut files = vec![readme_path];
-        if config.additional_pages.is_some() {
-            files.extend(config.additional_pages.as_ref().unwrap())
-        }
+        let files = if let Some(mut pages) = config.additional_pages.clone() {
```
ashleygwilliams: same here

ashleygwilliams:
i think we should close this as the original code is slightly better than this change and i think it will become irrelevant with the larger changes we'll be making soon

PR #52, opened 2022-12-26
Add social card config options

SaraVieira:
- Adds ability to pass a meta share image
- Ability to pass an alt to that image
- Ability to pass twitter id

### DIFF
--- a/oranda.json
+++ b/oranda.json
@@ -2,5 +2,10 @@
   "theme": "axo",
   "homepage": "https://oranda.axo.dev",
   "repository": "https://github.com/axodotdev/oranda",
-  "additional_pages": ["./docs/src/theme/themes.md"]
+  "additional_pages": ["./docs/src/theme/themes.md"],
+  "social": {
+    "image": "https://letsplayretro.games/meta.png",
+    "image_alt": "lOL",
+    "twitter_account": "@NikkitaFTW"
+  }
 }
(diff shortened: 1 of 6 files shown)


oranda.json
```diff
@@ -2,5 +2,10 @@
   "theme": "axo",
   "homepage": "https://oranda.axo.dev",
   "repository": "https://github.com/axodotdev/oranda",
-  "additional_pages": ["./docs/src/theme/themes.md"]
+  "additional_pages": ["./docs/src/theme/themes.md"],
+  "social": {
+    "image": "https://www.axo.dev/meta_small.jpeg",
+    "image_alt": "AXO",
```
ashleygwilliams: 😱 all caps?? lol

src/site/html.rs
```diff
@@ -2,6 +2,37 @@ use axohtml::{dom::DOMTree, html, text, unsafe_text};
 
 use crate::config::{theme, Config};
 use axohtml::elements::div;
+use axohtml::elements::meta;
+
+// Clippy issue as explained here
```
ashleygwilliams: I might say "false positive duplicate allocation warning" and link to this: https://github.com/rust-lang/rust-clippy/issues?q=is%3Aissue+redundant_allocation, which will be more inclusive of the types of errors (that issue is not *exactly* the one we have, but similar)

ashleygwilliams (approved):
would like a few *extremely* small changes but i think this is good to go otherwise!

src/site/html.rs
SaraVieira: Done!

oranda.json
SaraVieira: also fixed lol

PR #54, opened 2023-01-07
[DO NOT MERGE] run 'cargo-dist init --ci=github'

Gankra:


.github/workflows/release.yml
```diff
@@ -0,0 +1,127 @@
...
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Rust
+        run: rustup update stable && rustup default stable
+      - name: Install cargo-dist
+        # Currently we install cargo-dist from git, in the future when it's
+        # published on crates.io or has prebuilt binaries, we'll do better.
+        run: cargo install --git https://github.com/axodotdev/cargo-dist/
```
Gankra, line 77: this part needs to be changed before landing, we should be fetching prebuilt binaries from a public url (ideally with a `curl | sh` setup).

PR #56, opened 2023-01-09
Rename axo theme & make it default

SaraVieira:
- Renames `axo` theme to `axo-dark`
- Renames files and classes
- Makes theme default


SaraVieira:
fucked up git

PR #57, opened 2023-01-09
rename axo theme, make it default

### LINKED ISSUE #55: Make `axo-dark` default theme
Currently `light` is the default theme, but as we discussed some time ago, it would be better for us to default to our theme. I recommend we call this `axo-dark`, since we do have a light variation, which we can add as well, but maybe as step 2.

SaraVieira:
- Renames `axo` theme to `axo-dark`
- Renames files and classes
- Makes theme default

closes https://github.com/axodotdev/oranda/issues/55

### DIFF
--- a/oranda.json
+++ b/oranda.json
@@ -1,5 +1,4 @@
 {
-  "theme": "axo",
   "homepage": "https://oranda.axo.dev",
   "repository": "https://github.com/axodotdev/oranda",
   "additional_pages": ["./docs/src/theme/themes.md"],
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -127,7 +127,7 @@ impl Default for Config {
             name: String::from("My Axo project"),
             no_header: false,
             readme_path: String::from("README.md"),
-            theme: Theme::Light,
+            theme: Theme::AxoDark,
             remote_styles: vec![],
             additional_css: String::from(""),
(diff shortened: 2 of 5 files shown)


sarahghp:
Thanks for jumping on this, @SaraVieira. Can we also check if a theme is a dark theme and add the `dark` class to the body? That way it will work with the generated styles as in other apps.

I did it like this in `html.rs` to test, but it's not conditional, etc.

Screen Shot 2023-01-10 at 12 41 28 PM

sarahghp:
(Also I wonder why all the jobs are failing.)

SaraVieira:
For the classes that's already done automatically, you can check hy changing the theme 

The jobs are not starting so I'm assuming we ran out of minutes

sarahghp:
I does not apply `dark` for the `axo-dark` theme. Since we are pulling the classes from the generated Tailwind classes, it needs to do so. If it is somewhere else, you can update it elsewhere, but it needs to add it for axo dark.

SaraVieira:
I understand what you mean now, it should have dark and axo as classes

SaraVieira:
Done! Let me know if it makes sense

- light -> .light
- dark -> .dark
- axo-dark -> .dark.axo
- axo-light -> .light.axo

sarahghp (approved):

PR #61, opened 2023-01-10
Start docs

SaraVieira:
These docs are just scaffolding for the options oranda has, it will need plenty of changes but seems like a good start to keep going in docs

Would love opinions

Also adds docs for things not finished:

- Logo
- Analytics
- Axo Light theme


ashleygwilliams (approved):
merging despite having many comments, this is just to get started :)

PR #62, opened 2023-01-11
Read and sanitize raw html

### LINKED ISSUE #58: raw HTML is being omittted
Running with the `oranda.js` markdown file, it seems the images are not being rendered and are in fact being omitted. This is the rendered HTML from the first cell of the contributors table, which is generated from the following Markdown:

```md
| [<img src="https://avatars0.githubusercontent.com/u/1051509?v=4" width="100px;"/><br /><sub><b>Sara Vieira</b></sub>](http://iamsaravieira.com)<br />
[💻](https://github.com/axodotdev/oranda/commits?author=SaraVieira "Code") 
[🎨](#design-SaraVieira "Design") [🤔](#ideas-SaraVieira "Ideas, Planning, & Feedback") |
```
Screen Shot 2023-01-10 at 2 22 27 PM

(As a sidenote, these are rendered as all `<th>` elements, but they are actually cells and should be `<td>`.)

SaraVieira:
This PR fixes an issue where we were not reading raw HTML that was passed in the markdown. It also adds a [sanitization library](https://github.com/rust-ammonia/ammonia) for the same HTML to make sure nothing dangerous is passed.

As a sidenote, it seems all the things are rendered as `th` because `all-contributors` was not generating the tables right: https://github.github.com/gfm/#tables-extension-

Closes https://github.com/axodotdev/oranda/issues/58

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,6 +29,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "ammonia"
+version = "3.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "64e6d1c7838db705c9b756557ee27c384ce695a1c51a6fe528784cb1c6840170"
+dependencies = [
+ "html5ever",
+ "maplit",
+ "once_cell",
+ "tendril",
+ "url",
+]
+
+[[package]]
 name = "ansi_term"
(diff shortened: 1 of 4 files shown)


sarahghp (approved):
This looks good to me. Thank you for jumping on it!

PR #64, opened 2023-01-12
run cargo dist init --ci=github --installer=github-shell --installer=github-powershell

Gankra:
On paper this Just Works when you run

```
cargo release 0.1.0
```


Gankra:
(this should be rerun with the lastest version of cargo-dist)

ashleygwilliams:
@Gankra i'm not super clear on the status here- i think for now this looks good to merge if that is the intent

ashleygwilliams (approved):

ashleygwilliams:
(marking WIP until reply from gankra)

PR #66, opened 2023-01-13
fix code styles

SaraVieira:
by default amonia seems to remove all style tags


sarahghp (approved):
A custom element, blockquote, and styled pre blocks co-existing! Approved & merging.

Screen Shot 2023-01-16 at 11 58 43 AM

Issue #59, opened 2023-01-10
CSS for Oranda: Umbrella Issue


sarahghp:
I took some time to think through a system for CSS in Oranda that can use Fringe as the single source of truth for Axo styles and can be extended once we start shipping Sunfish and can work with user code. Ultimately, what I see us doing is:
* pulling the Axo themes from Fringe as remote assets
* allowing user files that can overwrite our themes, probably using an explicit [cascade layers declaration](https://css-tricks.com/css-cascade-layers/), rather than a specificity arms-race 
* using [`binary install`](https://github.com/axodotdev/binary-install/blob/main/src/lib.rs) to process files with the [Tailwind standalone CLI ](https://tailwindcss.com/blog/standalone-cli)so user files and internal, non-Axo themes can use `@apply` within Oranda and we can define Oranda-only components using these classes (like the header) — [related](https://github.com/rustwasm/wasm-pack/tree/master/src/install)
* extending the cascade layers override approach to themes and overrides from Sunfish
* adding class handles that users can use to target specific components (including Oranda components like the header), so they do not find themselves writing strange selecters, which I have definitely done with configurable, cloud-hosted blogs and the like

To get there, I have broken up what we need to do into a few groups, below. Essentially the priority is: 
1. display Fringe-generated styles; 
2. make it easy for users to write their own files
3. fill in the middle for Sunfish files and easier targetting for users

Here it is in rough milestones ⤵️

---

For v1 I intend to:

* Create an action that generates a stylesheet using the Tailwind CLI via an action and puts that stylesheet into our npm package, so we can reference it as a remote asset from Fringe via unpkg. I have a rough version of it done on the `generate-static-file` branch, but it is going to require some changes because the CLI command only works with CommonJS files.  

* Make it so that v1 includes only `axo-light` and `axo-dark`, which it can pull from Fringe, and otherwise only include CSS _within_ Oranda that is need for e.g. the header. This will be removed in v2.

* Update what needs updating to look refined (see: https://github.com/axodotdev/oranda/issues/60)

---

Then after that ships, for v2, we can:

* Use `binary install` to use the Tailwind standalone to process internal files and user files
* Returning other themes (this gives us something new to announce as well and me more time to review them, which is nice)

---

Then v2+:

* Figure out how to add handle classes, either while iterating through the `comrak` AST or using a tool that already implements a visitor patter. (I seem to recall us discussing one at one point, but this will need more research.)

---

[...]

PR #67, opened 2023-01-16
CSS layer update

### LINKED ISSUE #60: Oranda CSS fixes for v1
I finally took a look at the `oranda-js` README as generated by this version and this list captures some apparent fixes. 

My recommendation for how to attack these is not to include the Sass-generated `styles.css` for `axo-dark` and instead use just the Fringe-generated file. Styles can be added using @apply in Fringe and they will always be included, I believe.  (See https://github.com/axodotdev/oranda/issues/59 for more on that).

## Oranda Changes

Code styles (heavy red, for instance), spacing above & below code blocks

Screen Shot 2023-01-10 at 4 34 37 PM

---

List styles — we should probably have decorations for at least non-emoji lists and also indent. This may exist in blog styles. Also maybe slightly smaller size typeface than text?

Screen Shot 2023-01-10 at 4 44 08 PM
Screen Shot 2023-01-10 at 4 44 14 PM

---

Table styles: color (text white, header not orange — pink or purple), text vs. link sizing

Screen Shot 2023-01-10 at 4 56 48 PM

---

Type styles: h2, h4, should be pruple ... this should hopefully fall out from using only the Fringe styles

---

Image styles: no comment yet, not currently loading, see: https://github.com/axodotdev/oranda/issues/58

---

## Fringe Changes

* We should bring down the size of the h1 & h2, and add an h1-xl for the front page of the marketing site

sarahghp:
This PR updates Oranda to match the new CSS approach as outlined in https://github.com/axodotdev/oranda/issues/59. It is the pair of https://github.com/axodotdev/turbo-axo/pull/175 and relies on https://github.com/axodotdev/turbo-axo/pull/176 being released and published to work fully.

To see what this looks like locally, replace the `README` with [the oranda-js version](https://raw.githubusercontent.com/axodotdev/oranda-js/main/Readme.md). Creating a test file that can exist in this repo is on my list.

A few notes about the new CSS approach and its expansion into next steps; these are also mostly repeated in https://github.com/axodotdev/oranda/issues/59.

* From this PR, we are loading the themes as a remote stylesheet hosted on `unpkg` (from our fringe npm package). For the next step, we'll need to make this dynamic.

* This stylesheet defines three layers: `@base` (our standard fringe styles), `@theme` (the styles we provide), and `@author`, which we do not define. This last layer can be defined in `axo-asset` and will then take precedence over other layers automatically. (Bare CSS without a layer has the highest precedence.) We should include this in our docs and encourage user-defined themes, whenever they exist to implement the `@author` layer.

* Currently, we rely on the theme name to apply `dark` or not to `html` to leverage tailwind class-based dark theme toggling. We will want to account for that when users provide us remote assets. 

* This approach completely removes `scss`. It barfs on `@layer` directives (because they are new, I assume) and is therefore incompatible at a base level with the way we are using these layers to manage precedence. [More about layers ↬](https://css-tricks.com/css-cascade-layers/).


src/site/html.rs
```diff
@@ -58,12 +58,13 @@ pub fn build(config: &Config, content: String) -> String {
     <meta property="og:type" content="website" />
     <meta property="og:title" content=&config.name />
     {social_meta}
+    <link rel="stylesheet" href="http://sarahghp.com/themes/axo-oranda.css"></link>
```
sarahghp: Update to  `https://www.unpkg.com/browse/@axodotdev/fringe/themes/axo-oranda.css` once https://github.com/axodotdev/turbo-axo/pull/176 is merged and released.

src/site/mod.rs
```diff
@@ -13,6 +13,7 @@ use crate::config::theme::Theme;
 
 use crate::config::Config;
 
+#[derive(Debug)]
```
ashleygwilliams, line 16: nice

src/site/mod.rs
```diff
@@ -111,11 +110,11 @@ fn reads_description() {
...
     let site = Site::build(&config(), &config().readme_path).unwrap();
-    assert!(site.html.contains("<div class=\"body dark\">"));
+    assert!(site.html.contains("html class=\"dark\""));
 }
 
 #[test]
 fn reads_additional_css() {
     let site = Site::build(&config(), &config().readme_path).unwrap();
-    assert!(site.css.contains("#oranda body{background:red}"));
+    assert!(site.css.contains("body {\n  background: red;\n}\n"));
```
ashleygwilliams: oof this whitespace makes me worried

ashleygwilliams (approved):
am slightly worried about the white space and exact match in the css test, but i was also worried before. overall tho i think this looks good!

SaraVieira:
🎉

PR #72, opened 2023-01-18
Output

ashleygwilliams:
Screen Shot 2023-01-23 at 5 55 05 PM
Screen Shot 2023-01-23 at 5 54 57 PM

todo:
- [ ] switch println to something with log levels
- [x] only use colors if colors are supported

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 public/
+oranda-debug.log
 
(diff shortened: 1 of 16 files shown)


ashleygwilliams:
closing in favor of https://github.com/axodotdev/oranda/pull/108 because friggin conflicts are too much, don't wanna deal

PR #73, opened 2023-01-18
Add ability to pass a favicon

SaraVieira:

### DIFF
--- a/src/site/head.rs
+++ b/src/site/head.rs
@@ -37,10 +37,7 @@ pub fn get_favicon(favicon: String, dist_dir: String) -> Result<Box<link<String>
     let copy_result_future = axoasset::copy(&favicon, &dist_dir[..]);
     let copy_result = tokio::runtime::Handle::current().block_on(copy_result_future)?;
 
-    let path_as_string = match copy_result.strip_prefix(dist_dir) {
-        Ok(path) => Ok(path.to_string_lossy()),
-        Err(e) => Err(OrandaError::Other(e.to_string())),
-    }?;
+    let path_as_string = copy_result.strip_prefix(dist_dir)?.to_string_lossy();
 
     Ok(html!(<link rel="icon" href=path_as_string />))
 }
--- a/src/site/head.rs
+++ b/src/site/head.rs
@@ -45,6 +45,9 @@ pub fn get_favicon(favicon: String, dist_dir: String) -> Result<Box<link<String>
     Ok(html!(<link rel="icon" href=path_as_string />))
 }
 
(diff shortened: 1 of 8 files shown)


src/site/head.rs
```diff
@@ -0,0 +1,67 @@
+use crate::config::Config;
```
ashleygwilliams, line 1: i think it would be good to make all these functions private and then wrap them in a public function called "build" that return all the html for the head- thoughts?

src/site/header.rs
```diff
@@ -17,10 +17,10 @@ async fn fetch_logo(
 ) -> Result<Box<img<String>>> {
     let copy_result = axoasset::copy(&origin_path, dist_dir).await?;
 
-    let path_as_string = copy_result
-        .strip_prefix(dist_dir)
-        .unwrap()
-        .to_string_lossy();
+    let path_as_string = match copy_result.strip_prefix(dist_dir) {
+        Ok(path) => Ok(path.to_string_lossy()),
+        Err(e) => Err(OrandaError::Other(e.to_string())),
```
ashleygwilliams: what is this error- could we use a ? instead?

ashleygwilliams (changes requested):
a few small things!

src/site/header.rs
SaraVieira: changed, added the `StripPrefixError` to our errors

ashleygwilliams: danke!

src/site/head.rs
SaraVieira: That was my first thought but unfortunately I got several errors, first I attempted to return all the html inside the head, and it gave me an error because the HTML I was returning didn't have a parent element and then when I added this element I got a new error for axohtml

CleanShot 2023-01-18 at 14 47 45@2x

SaraVieira: Error comes from here: https://github.com/axodotdev/axohtml/blob/remove-dodrio/macros/src/html.rs#L31

ashleygwilliams: let's file an issue and get this PR merged in the meantime? i'll need more info to debug this properly

SaraVieira: Sure! Can I get an approve to merge?

ashleygwilliams: issue filed https://github.com/axodotdev/oranda/issues/74

ashleygwilliams (approved):

SaraVieira:
Issue created here: https://github.com/axodotdev/oranda/issues/75

PR #76, opened 2023-01-18
Merge css options & use axo asset

SaraVieira:
## User facing
- Uses axo asset for fetching css
- Removes the css:build functions as we just copy them over
- merges both options accepting a vec now

## Tests
- Creates a laxy static for the test enviroment and uses that everywhere
- deletes shared initioalized tokio function
- moves most tests to the same file in tests/mod.rs


SaraVieira:
I fixed the tests butr it's really cursed, would love a better idea

cc @Gankra @ashleygwilliams

SaraVieira:
updates on this curse, fixing the tests broke remote assets: https://github.com/axodotdev/oranda/actions/runs/3951628662/jobs/6765726311

By moving the runtime creation inside the main it fixes it and honestly I have no idea
[`1520621` (#76)](https://github.com/axodotdev/oranda/pull/76/commits/1520621baa847eb70c9da173dc91023f7bba34fe)

src/errors.rs
```diff
@@ -25,9 +25,6 @@ pub enum OrandaError {
     #[error("failed to read {filedesc} at {path}")]
     FileNotFound { filedesc: String, path: String },
 
-    #[error("failed to fetch {resource} at {url}")]
-    RequestFailed { resource: String, url: String },
-
```
SaraVieira, line 33: Was not used anymore

ashleygwilliams: nice!

src/main.rs
```diff
@@ -21,6 +22,16 @@ enum Command {
     Serve(Serve),
 }
 
+pub fn create_runtime_tokio() -> Runtime {
```
ashleygwilliams: i'm slightly confused about this function- and then the contents of this function are repeated in the main function. can you tell me more? or we can hop on a call

ashleygwilliams: also you can probably skip the assignment here (remove the `let runtime` and `runtime`)

src/site/mod.rs
```diff
@@ -92,21 +77,23 @@ fn config() -> Config {
...
-        additional_css: String::from("./src/site/fixtures/additional.css"),
+        additional_css: vec![String::from("./src/site/fixtures/additional.css")],
         theme: Theme::Dark,
         ..Default::default()
     }
 }
 
 #[test]
 fn it_builds_the_site() {
+    let _guard = create_runtime_tokio().enter();
```
ashleygwilliams: oh! i see.

ashleygwilliams (commented):
i am not entirely sure if this is cursed per se, but i see why the unit tests needing to spin up their own runtime can feel like "a lot". i do wonder if there is a better way.

src/main.rs
SaraVieira: that was my comment in the top, when using this function in main the remote assets broke, genuinly don't understand

src/site/html.rs
```diff
@@ -32,15 +32,17 @@ pub fn build(config: &Config, content: String) -> Result<String> {
...
     let doc: DOMTree<String> = html!(
     <html lang="en" id="oranda" class=theme>
         <head>
             <title>{ text!(&config.name) }</title>
             {homepage}
             {favicon}
             {meta_tags}
-            <link rel="stylesheet" href="https://www.unpkg.com/@axodotdev/fringe/themes/axo-oranda.css"></link>
-            <link rel="stylesheet" href="styles.css"></link>
+            <link rel="stylesheet" href="https://www.unpkg.com/@axodotdev/fringe@0.0.7/themes/axo-oranda.css"></link>
```
ashleygwilliams: i thought we were planning on treating fringe the same as the other css and vendoring it?

src/site/mod.rs
```diff
@@ -115,18 +105,14 @@ fn reads_description() {
 
 #[test]
 fn reads_theme() {
+    let _guard = initialize_tokio_runtime().enter();
     let site = Site::build(&config(), &config().readme_path).unwrap();
     assert!(site.html.contains("html class=\"dark\""));
 }
 
-#[test]
```
ashleygwilliams, line 131: can you say more about deleting this?

SaraVieira: We don't build css anymore, we just copy the files

src/site/html.rs
SaraVieira: I tried that, we need to do some more stuff since when copying it the fringe file requires other files that are not present

ashleygwilliams: oh hrmmm so the fringe file here makes requests to other files?

src/site/mod.rs
ashleygwilliams: ah- should we add a test for that then?

src/site/html.rs
SaraVieira: yeah it requires `finge-output` & `theme.css`

ashleygwilliams: ok can you open an issue about this and then we can just leave this for now- i may be wrong on this but isn't css that requests other css often a performance issue?

src/main.rs
```diff
@@ -4,11 +4,10 @@ mod commands;
 mod config;
 mod errors;
 mod site;
+pub mod tests;
```
ashleygwilliams, line 7: i don't think we should need to do this- i think we want to lift the tests dir out of the src dir and into the top level :)

ashleygwilliams (approved):
we should move the tests dir but otherwise i think this is awesome!! nice goin' 🚀

src/main.rs
SaraVieira: If I get it in the main dir I thgen can't import stuff from the crate :/

What do you mean?

ashleygwilliams: ah- so usually a crate has a top level tests directory for testing. the tests/mod file does have tests in it tho- just test utils. i think in general this stuff could be re-org'd but we can do it in a diff PR. i'll open an issue

PR #79, opened 2023-01-19
Allow passing a config file path

### LINKED ISSUE #49: Remove extraneous call to `Config::build`
Right now in https://github.com/axodotdev/oranda/blob/main/src/site/markdown/syntax_highlight/mod.rs#L32 we are calling `Config::build` and we should not call that more than once for performance and code cleanness reasons

### LINKED ISSUE #70: parameterize oranda config file path
We currently assume it is always in root, but we should make it so its possible to configure the filepath as well.

### LINKED ISSUE #78: don't sanitize classes out of raw markdown
When working on https://github.com/axodotdev/oranda/pull/77, I found this line:

```
<div class="title text-slate-500 dark:text-slate-50">For instance, this is a plain div with a class attached to style it in the title style. <sup>Wow, is it big.</sup></div>

```

is presented without its classes when built.

SaraVieira:
This PR allows the user to pass a path to a config file if they don't wish to use `oranda.json`

closes https://github.com/axodotdev/oranda/issues/49

closes https://github.com/axodotdev/oranda/issues/70

closes https://github.com/axodotdev/oranda/issues/78

### DIFF
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -10,11 +10,13 @@ use crate::site::Site;
 pub struct Build {
     #[arg(long, default_value = "./")]
     path: PathBuf,
+    #[arg(long, default_value = "./oranda.json")]
+    config: PathBuf,
 }
 
 impl Build {
     pub fn run(&self) -> Result<()> {
-        let config = Config::build()?;
+        let config = Config::build(&self.config)?;
         Site::write(&config)?;
         Ok(())
     }
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -9,14 +9,14 @@ use crate::site::Site;
 #[derive(Debug, Parser)]
 pub struct Build {
     #[arg(long, default_value = "./")]
-    path: PathBuf,
+    project_root: PathBuf,
     #[arg(long, default_value = "./oranda.json")]
-    config: PathBuf,
+    config_path: PathBuf,
 }
 
 impl Build {
     pub fn run(&self) -> Result<()> {
-        let config = Config::build(&self.config)?;
+        let config = Config::build(&self.config_path)?;
         Site::write(&config)?;
         Ok(())
     }
--- a/src/config/oranda.rs
+++ b/src/config/oranda.rs
@@ -36,12 +36,10 @@ pub struct OrandaConfig {
 
 impl OrandaConfig {
     pub fn load(config_path: &Path) -> Result<Option<OrandaConfig>> {
-        println!("reading from oranda config...");
         let config_future = axoasset::load_string(config_path.to_str().unwrap());
 
         let config = tokio::runtime::Handle::current().block_on(config_future)?;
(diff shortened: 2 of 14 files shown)


src/site/markdown/mod.rs
```diff
@@ -56,17 +60,17 @@ fn load(readme_path: &Path) -> Result<String> {
...
 
     let mut plugins = ComrakPlugins::default();
-    let adapter = Adapters {};
+    let adapter = Adapters { syntax_theme };
     plugins.render.codefence_syntax_highlighter = Some(&adapter);
 
     let unsafe_html = comrak::markdown_to_html_with_plugins(&readme, &options, &plugins);
     let safe_html = Builder::new()
-        .add_generic_attributes(&["style"])
+        .add_generic_attributes(&["style", "class", "id"])
```
SaraVieira, line 73: this is a fix for: https://github.com/axodotdev/oranda/issues/78

src/commands/build.rs
```diff
@@ -10,11 +10,13 @@ use crate::site::Site;
 pub struct Build {
     #[arg(long, default_value = "./")]
     path: PathBuf,
```
ashleygwilliams: i wonder if these names will come back to bite us- maybe "project_root" and "config_path"?

src/config/oranda.rs
```diff
@@ -38,16 +35,13 @@ pub struct OrandaConfig {
...
-            println!("read data: {:?}", &data);
-            Ok(Some(data))
-        } else {
-            Ok(None)
-        }
+        let config_future = axoasset::load_string(config_path.to_str().unwrap());
+
+        let config = tokio::runtime::Handle::current().block_on(config_future)?;
+        let data: OrandaConfig = serde_json::from_str(config.as_str())?;
+        println!("read data: {:?}", &data);
```
ashleygwilliams: did you mean to keep this?

src/site/markdown/mod.rs
```diff
@@ -9,10 +9,14 @@ use std::collections::HashMap;
 use std::fs;
 use std::path::Path;
 
-pub struct Adapters {}
-impl SyntaxHighlighterAdapter for Adapters {
+use self::syntax_highlight::syntax_themes::SyntaxThemes;
+
+pub struct Adapters<'a> {
+    syntax_theme: &'a SyntaxThemes,
```
ashleygwilliams: we might want to rename the enum to the singular `SyntaxTheme`

ashleygwilliams (approved):
extremely small comments

src/config/oranda.rs
SaraVieira: deleted this and also the ones from the js and rust config files

src/site/markdown/mod.rs
SaraVieira: done!

src/commands/build.rs
SaraVieira: done!

PR #81, opened 2023-01-20
replace readme with readme path in dev

SaraVieira:
with this we can now run 

```
cargo run -- build --config-path=dev/oranda.json
```

or with watch: 

```
cargo watch -c -w src -x  'run -- build --config-path=dev/oranda.json'
```

and get a dev version

### DIFF
--- a/dev/oranda.json
+++ b/dev/oranda.json
@@ -1,6 +1,6 @@
 {
   "homepage": "https://oranda.axo.dev",
-  "readme": "dev/README.md",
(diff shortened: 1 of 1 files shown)


ashleygwilliams (approved):
this seems fine but i do not fully understand what this is doing? was it just that the config option wasn't renamed in a previous PR?

SaraVieira:
Yeah, i merged it without changing this, my bad

PR #82, opened 2023-01-20
Copy over static folder

### LINKED ISSUE #47: Copy over static folder
This issue focuses on copying any static assets the user has whether they are linked or not

For this we need:

- Add static_folder option and default it to "./static"
- Copy over the content of this folder to the dist directory

SaraVieira:
PR copies the static folder recursively

Maybe this functionlity could be implemented in axoasset

closes https://github.com/axodotdev/oranda/issues/47


src/site/mod.rs
```diff
@@ -8,6 +8,9 @@ mod head;
 mod header;
 mod html;
 pub mod markdown;
+extern crate fs_extra;
```
ashleygwilliams: why is this here? you shouldn't need this (it's a very old and totally vestigal piece of rust...)

src/site/mod.rs
```diff
@@ -8,6 +8,9 @@ mod head;
...
+use fs_extra::copy_items;
+use fs_extra::dir;
```
ashleygwilliams: i would not bring these into scope because it will be confusing where they come from

ashleygwilliams (approved):
so this seems OK, i do think we'll want to put this in axo asset but it's fine in here for now. a little worried about this dep - but i think for now we can keep it and revisit when we move to axoasset- can you file an issue on axoasset for this?

src/site/mod.rs
SaraVieira: you are right, its removed

src/site/mod.rs
SaraVieira: also fixed

PR #83, opened 2023-01-20
add path_prefix

SaraVieira:
Adds the option to pass a path_prefix so that websites like github pages can work, right now we only support websites at the root and links will not work like so: https://axodotdev.github.io/oranda/

This fixes that 


ashleygwilliams (approved):

docs/src/gh-pages.md
```diff
@@ -0,0 +1,13 @@
+# Hosting on Github Pages
+
+So you have set up oranda, set up some links and deployed it to Github pages with CI and now the links won't work?
```
ashleygwilliams: so, one thing that's worth noting is that not all github pages deployments are not at the root. personal/org gh pages (e.g. axodotdev.github.io) are actually at the root. so it may be useful to identify this particular scenario with the example (your line 7) first instead of assuming they are already at the failure state.

docs/src/options/path_prefix.md
```diff
@@ -0,0 +1,11 @@
+# Path Prefix
+
+Sometimes your website may not be hosted at the root at of a domain and your nav links will not be point to the right place.
```
ashleygwilliams: this is either missing a word or has an extra one :)

ashleygwilliams (changes requested):

docs/src/options/path_prefix.md
SaraVieira: how did I write a book ashley wat??

docs/src/gh-pages.md
SaraVieira: fixed

ashleygwilliams (approved):
🚀 wooo

PR #84, opened 2023-01-21
Join head metas in one file

SaraVieira:
NOT WORKING


ashleygwilliams:
i have an idea for this that doesn't require an axohtml change as part of #135 (which will be translated into a series of smaller PRs) - closing this in favor of that as this is gonna run into a lot of merge conflicts soon

PR #94, opened 2023-01-23
allow prefix in oranda serve

### LINKED ISSUE #71: Add oranda theme dev
Add command to build & serve oranda with a link to a theme config that includes all possible items (repo bar, multiple pages, README with markdown elements). Predicated on https://github.com/axodotdev/oranda/issues/70

### LINKED ISSUE #90: oranda serve does not play nicely with `pathprefix`
pages

SaraVieira:
closes https://github.com/axodotdev/oranda/issues/90

closes https://github.com/axodotdev/oranda/issues/71

### DIFF
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -1,10 +1,10 @@
-use std::net::SocketAddr;
-use std::path::PathBuf;
-
 use crate::config::Config;
 use crate::errors::*;
 use axum::{http::StatusCode, routing::get_service, Router};
+use axum_extra::routing::SpaRouter;
 use clap::Parser;
+use std::net::SocketAddr;
+use std::path::PathBuf;
 use tower_http::services::ServeDir;
 
 #[derive(Debug, Parser)]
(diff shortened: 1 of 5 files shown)


src/site/header.rs
```diff
@@ -41,7 +41,7 @@ pub fn create(config: &Config) -> Result<Box<header<String>>> {
                 let mut href = format!("/{}", file_name);
 
                 if let Some(prefix) = &config.path_prefix {
-                    href = format!("/{}{}", prefix, href);
```
SaraVieira, line 44: This seems to be a fix but it's super cumbersome

https://github.com/tokio-rs/axum/discussions/446

ashleygwilliams: oh interesting, def a bummer. might be a nice opportunity to contribute back. regardless maybe we can also snag some of the error work from that answer so we don't just serve the empty page if there's an issue?

src/commands/serve.rs
```diff
@@ -33,10 +33,20 @@ impl Serve {
             },
         );
 
-        let app = Router::new().fallback(static_files_service);
+        let app;
+        let route;
+        if let Some(path_prefix) = config.path_prefix {
+            route = format!("/{}", path_prefix);
```
ashleygwilliams: is the slash necessary? my instinct is it wouldn't be but i'm not sure

ashleygwilliams: i think this is a scenario where we may want to use `Url` https://docs.rs/url/latest/url/struct.Url.html#method.join

ashleygwilliams (approved):
small requests- if you don't do the error handling in this one we should open an issue for it

src/commands/serve.rs
SaraVieira: oh no

```
thread 'main' panicked at 'Paths must start with a `/`', src/commands/serve.rs:41:18
```

ashleygwilliams: well that settles that ha

src/site/header.rs
SaraVieira: We still need this right?

ashleygwilliams: no i dont think so, i ran it locally and it worked?

SaraVieira:
Can't approve because it's my PR but ship it!

PR #96, opened 2023-01-24
Adds the last elements to styles dev

sarahghp:
Not all of these are styled correctly, but I think it's a good start. Fringe PR to follow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 /target
 public/
-
+**/.DS_Store
--- a/dev/README.md
+++ b/dev/README.md
@@ -1,11 +1,13 @@
+<!-- This is a comment and doesn't appear -->
+
 > This README contains expected Markdown elements. It is useful for developing themes and testing that changes to Oranda parsing do not have unexpected repercussions. If you run into a bug not captured here, please add it.
 
 # Heading 1
 ## Heading 2
 ### Heading 3
 #### Heading 4
-###### Heading 5
-####### Heading 6
+##### Heading 5
+###### Heading 6
 
 A paragraph can contain **bold**, _italic_ and *italic* text. (Those two are different.) Let's look at some longer text and other features.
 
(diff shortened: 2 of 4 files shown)


dev/oranda.json
```diff
@@ -1,6 +1,7 @@
 {
   "homepage": "https://oranda.axo.dev",
   "readme_path": "dev/README.md",
+  "theme": "light",
```
SaraVieira, line 4: Would you mind defaulting to black?

sarahghp: So the reason I did this is to make sure when people are testing dev stuff, they don't forget to test the theme option and they don't forget to consider light mode. I would prefer not to change it, but I am open to reasons I haven't considered.

dev/README.md
```diff
@@ -50,6 +52,9 @@ One error we ran into before was in code-blocks with HTML specifically, so let's
 ![An image from a vintage video synth](./static/some_art.png)
 Here is an image I made.
 
+Here is one with an external link
+![](https://sarahghp.com/vids/noodling/noodl06)
```
sarahghp: Also, I almost forgot, we should link to something that is not mine. Do you have a suggestion for a stable remote link?

SaraVieira: We can use unsplash images, they don't get deleted

https://unsplash.com/photos/J4CeJunkE8M/download?ixid=MnwxMjA3fDB8MXxhbGx8MTR8fHx8fHwyfHwxNjc0NTYzNDA0&force=true&w=640

dev/oranda.json
SaraVieira: oh this is the dev yeah sorry

dev/README.md
sarahghp: Updated with pexels, since it is rights free. Thank you for the suggestion!

.gitignore
```diff
@@ -1,3 +1,3 @@
 /target
 public/
-
+**/.DS_Store
```
ashleygwilliams, line 3: this is an OS specific file and should be in a global gitignore, not a project gitignore

dev/additional.md
```diff
@@ -1,66 +1,13 @@
...
-brown and white in trees,
-snows and skies of laughter
-always diminishing, less funny
-not just darker, not just grey.
-
-It may be the coldest day of
-the year, what does he think of
-that? I mean, what do I? And if I do,
-perhaps I am myself again.
+<h2 style="margin-bottom:0">Snail</h2>
```
ashleygwilliams, line 1: is this inline style on purpose?

SaraVieira (approved):

ashleygwilliams (changes requested):

PR #98, opened 2023-01-24
Fix console warning

### LINKED ISSUE #88: console warning
Screen Shot 2023-01-21 at 8 38 15 AM

i am not 100% sure what this means, but we should aim for oranda pages to have a clean console to the extent that is within our control

SaraVieira:
closes https://github.com/axodotdev/oranda/issues/88

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -119,9 +119,9 @@ dependencies = [
 
 [[package]]
 name = "axohtml"
-version = "0.3.0"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f3ea0f7de9bc9c63e749dbc6adc0cf2000dcd59d7de8327ec26ee23bd5a6a542"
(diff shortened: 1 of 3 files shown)


ashleygwilliams (approved):
would be useful to double check this does indeed get rid of the error, but assuming it does, this looks good

SaraVieira:
We can really only check when merged, no? :/

SaraVieira:
Tests are failing because a change in axo-html is required: https://github.com/axodotdev/axohtml/pull/23

ashleygwilliams:
axohtml is published, so i'm gonna snag this and get it over the hump

ashleygwilliams:
annnnnd axohtml needs a patch, 

```
---- site::it_builds_the_site stdout ----
ERROR: <meta http_equiv="Permissions-Policy"> failed to parse attribute value: Matching variant not found
ERROR: rebuild with nightly to print source location
thread 'site::it_builds_the_site' panicked at 'failed to parse string literal', src/site/head.rs:59:9
```

Issue #89, opened 2023-01-22
Create artifacts downloads


SaraVieira:
This will involves many steps:

### Common

- [x] Create system names so that they can be used everywhere
- [x] create detect-os js file
- [x] Show download link for correct system
- [ ] Create artifacts as page with table that includes all options

### Cargo dist

- [x] use https://github.com/axodotdev/cargo-dist/blob/v0.2.2-prerelease6/cargo-dist-schema/src/lib.rs to parse json
- [x] In cargo dist add curl
- [x] When cargo dist is true get manifest
- [x] windows downloads the ps1 installer
- [x] add link to sourcemap on windows
- [x] add installer hint to each system

### Binaries

- [ ] Allow hashmap where the keys must be a system type we return from detect-os
- [ ] Show correct download link
- [ ] Add link to artifacts created

### Type

```ts
type Distributables = {
	link_downloads: Bool
	// requires repo
	cargo_dist: true,
	package_managers: {
		[string]: ""
	},
	binaries: {
		linux: ""
	}
}
```

ashleygwilliams:
closed by #99

SaraVieira:
Reopened as we are missing binaries and package managers

ashleygwilliams:
is this closable now, @SaraVieira ?

ashleygwilliams:
closing as i believe this is complete.

PR #99, opened 2023-01-25
Add support for cargo dist

SaraVieira:
This PR adds support for cargo dist by showing a download button for the user's OS, for this it does a couple of things

- Adds OS detection & shows correct one in the same script
- Adds support for reading version and repo from the cargo.toml
- Creates new page with table showing all links and downloads

related https://github.com/axodotdev/oranda/issues/89


src/config/artifacts.rs
```diff
@@ -0,0 +1,154 @@
...
+    for release in manifest.releases.iter() {
+        for artifact in release.artifacts.iter() {
+            let name = &artifact.name;
+            let url = create_download_link(config, name);
+            table.push(html!(
+            <tr>
+                // <td>{name}</td>
+                // <td>{artifact.kind}</td>
+                <td><a href=url>{text!("Download")}</a></td>
+            </tr>
```
SaraVieira: If I uncomment any of these lines I get this massive error

```
error[E0599]: the method `into_iter` exists for reference `&std::string::String`, but its trait bounds were not satisfied
   --> src/config/artifacts.rs:139:24
    |
139 |               table.push(html!(
    |  ________________________^
140 | |             <tr>
141 | |                  <td>{name}</td>
142 | |                 // <td>{artifact.kind}</td>
143 | |                 <td><a href=url>{text!("Download")}</a></td>
144 | |             </tr>
145 | |             ));
    | |_____________^
    |
   ::: /Users/saravieira/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/src/rust/library/alloc/src/string.rs:367:1
    |
367 |   pub struct String {
    |   -----------------
    |   |
    |   doesn't satisfy `std::string::String: IntoIterator`
    |   doesn't satisfy `std::string::String: Iterator`
    |
    = note: the following trait bounds were not satisfied:
            `&std::string::String: Iterator`
            which is required by `&std::string::String: IntoIterator`
            `&&std::string::String: Iterator`
            which is required by `&&std::string::String: IntoIterator`
            `&mut &std::string::String: Iterator`
            which is required by `&mut &std::string::String: IntoIterator`
            `std::string::String: Iterator`
            which is required by `std::string::String: IntoIterator`
            `&mut std::string::String: Iterator`
            which is required by `&mut std::string::String: IntoIterator`
            `str: Sized`
            which is required by `str: IntoIterator`
            `str: Iterator`
            which is required by `str: IntoIterator`
            `&str: Iterator`
            which is required by `&str: IntoIterator`
            `&mut str: Iterator`
            which is required by `&mut str: IntoIterator`
    = note: this error originates in the macro `html` (in Nightly builds, run with -Z macro-backtrace for more info)
```

dev/oranda.json
```diff
@@ -1,8 +1,11 @@
 {
   "homepage": "https://oranda.axo.dev",
   "readme_path": "dev/README.md",
+  "repository": "https://github.com/axodotdev/cargo-dist",
+  "artifacts": {
+    "cargo_dist": true
+  },
```
SaraVieira: this is for testing since cargo dist has cargo dist and we do not, will be deleted when the PR is done

src/site/header.rs
```diff
@@ -65,3 +65,16 @@ pub fn create(config: &Config) -> Result<Box<header<String>>> {
         </header>
     ))
 }
+
+pub fn repo_banner(config: &Config) -> Option<Box<div<String>>> {
+    config.repository.as_ref().map(|repository| {
+        html!(
```
Gankra: the current impl is good, but here's another way to do it if you're interested:

```rust
let repository = config.respository.as_ref()?;
html!(....);
```

(the `?` operator works on Options if your function returns an Option)

src/site/html.rs
```diff
@@ -19,6 +20,10 @@ pub fn build(config: &Config, content: String) -> Result<String> {
         true => None,
         false => Some(header::create(config)?),
     };
+    let os_script = match config.artifacts {
+        None => None,
+        Some(_) => Some(artifacts::get_os_script(config)?),
+    };
```
Gankra, line 25: this would probably be cleaner as a `map` but idk if you're futureproofing

src/config/artifacts.rs
```diff
@@ -0,0 +1,154 @@
...
+    if let Some(repo) = &config.repository {
+        if let Some(version) = &config.version {
+            format!("{}/releases/download/v{}/{}", repo, version, name)
+        } else {
+            String::from("")
+        }
+    } else {
+        String::from("")
+    }
+}
```
Gankra: `String::from("")` is a more awkward form of `String::new()`. If you don't anticipate wanting different behaviour for these paths this could be

```
if let (Some(repo), Some(version)) = (&config.repository, &config.version) {
  format!(...);
} else {
  String::new()
}
```

Alternatively you could make this function return an Option and do:

```rust
let repo = config.repository.as_ref()?;
let version = config.version.as_ref()?;
Some(format!(...))
```

src/site/html.rs
```diff
@@ -45,25 +49,25 @@ pub fn build(config: &Config, content: String) -> Result<String> {
...
-        <div class="repo_banner">
-            <a href=repository>
-                <div class="icon" aria-hidden="true"/>
-                {text!("Check out our GitHub")}
-            </a>
-        </div>
-        )
-    })
+pub fn build(config: &Config, content: String) -> Result<String> {
+    let artifacts_tabs = create_artifacts_tabs(config).unwrap();
```
Gankra: Should this unwrap be here?

src/site/javascript/detect_os.js
```diff
@@ -0,0 +1,143 @@
+/* Code modified from the blender website
+ * https://www.blender.org/wp-content/themes/bthree/assets/js/get_os.js?x82196
+ */
+
```
Gankra, line 4: not looking closely at this one, I assume it's fine

src/config/mod.rs
```diff
@@ -105,17 +114,24 @@ impl Config {
                     theme: custom.theme.unwrap_or(default.theme),
                     remote_styles: custom.remote_styles.unwrap_or(default.remote_styles),
                     additional_css: custom.additional_css.unwrap_or(default.additional_css),
-                    repository: custom.repository,
+                    repository: Some(
+                        custom
+                            .repository
+                            .unwrap_or_else(|| project.repository.unwrap_or_default()),
+                    ),
```
Gankra: hmm do you really want to end up with `Some("")` if nothing is set? Why not let it be an actual Option?

```
repository: custom.respository.or(project.repository)
```

src/config/mod.rs
```diff
@@ -105,17 +114,24 @@ impl Config {
...
                     syntax_theme: custom.syntax_theme.unwrap_or(default.syntax_theme),
                     analytics: custom.analytics,
                     additional_pages: custom.additional_pages,
                     social: custom.social,
+                    artifacts: custom.artifacts,
+                    version: Some(project.version.unwrap_or_default()),
```
Gankra: again why not just let this be an Option? Empty strings aren't terribly useful values here.

src/config/artifacts.rs
Gankra: This is a baffling error but I'm blind-guessing it's mad that `artifact.kind` doesn't implement Display -- for now you should match on it and choose the strings you want the different values to become.

Gankra (commented):
So far the general gist seems to make sense?

src/config/artifacts.rs
SaraVieira: that was also my assumption but even leaving that one commented it errors out, I think it's a axohtml bug

src/site/header.rs
SaraVieira: Ohhh, being able to use `?` when you return an option changes everything

src/site/html.rs
SaraVieira: yeah, when it's artifacts but only package managers I don't actually wanna get the script

src/config/artifacts.rs
SaraVieira: YES! I was looking into a way to do something like that but didn't know how! Changed!

src/site/html.rs
SaraVieira: Nop, good catch, removed

src/config/mod.rs
SaraVieira: yeah you are super right, changed

src/config/mod.rs
SaraVieira: also changed

src/site/html.rs
```diff
@@ -43,31 +46,32 @@ pub fn build(config: &Config, content: String) -> Result<String> {
             {homepage}
             {favicon}
             {meta_tags}
-            {fringe_css}
+            // {fringe_css}
             {additional_css}
+            <link rel="stylesheet" href="http://localhost:1234/axo-oranda.css"></link>
```
SaraVieira: Clippy is because of this, will be removed before merge but rn it's better for testing

ashleygwilliams: we should come up with a way long term that this is a better dev exp

dev/oranda.json
ashleygwilliams: we can update this i think!

src/config/artifacts.rs
```diff
@@ -0,0 +1,183 @@
...
+use axohtml::elements::{div, script, span};
+use axohtml::{html, text, unsafe_text};
+use cargo_dist_schema::{ArtifactKind, DistManifest};
+use serde::Deserialize;
+use std::fs::File;
+use std::io::Write;
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "lowercase")]
+enum ArtifactSystem {
```
ashleygwilliams, line 5: i feel like this could be its own crate (don't need to do right now, but just noting it)

src/config/artifacts.rs
```diff
@@ -0,0 +1,183 @@
...
+
+fn create_download_link(config: &Config, name: &String) -> String {
+    if let (Some(repo), Some(version)) = (&config.repository, &config.version) {
+        format!("{}/releases/download/v{}/{}", repo, version, name)
+    } else {
+        String::new()
+    }
+}
+
+pub fn create_artifacts_tabs(config: &Config) -> Result<Option<Box<div<String>>>> {
```
ashleygwilliams: this is a big function may be valuable to try to break it into smaller ones if possible

ashleygwilliams (commented):
re: organization- if the module returns html we should have it in site, if it returns config(build input) we should put it in config- i think it'd be valuable to go through these files and try to separate that out- i can take a pass once everything is settled and/or we could pair

src/config/artifacts.rs
SaraVieira: ended up using div's

src/config/artifacts.rs
SaraVieira: Done, created one for the install_hint

src/config/artifacts.rs
SaraVieira: It should be but maybe in cargo dist no?

ashleygwilliams: yeah i agree with that, curious what you think @Gankra

ashleygwilliams (approved):
happening

PR #104, opened 2023-01-26
theme & CSS docs

sarahghp:
Feel free to move and change as needed. With luck this is a good base.


SaraVieira (approved):
This looks great!

docs/src/options/theme.md
```diff
@@ -1,18 +1,92 @@
 # Theme
 
-By default Oranda comes with four default themes.
+Oranda comes with two default themes:
 
 ## ADD IMAGES UNDER EACH
```
ashleygwilliams: should we hold off merging until we get images here?

ashleygwilliams: we could also publish 2 example pages to show off the themes

ashleygwilliams (approved):

ashleygwilliams (commented):
this is great! thank you!

PR #105, opened 2023-01-26
lookup paths relative to project root not cwd

### LINKED ISSUE #95: look for readme relative to oranda config not current working dir
while working on https://github.com/axodotdev/oranda/pull/72 i benefitted from reliably making oranda error when it could not find README.md because i called oranda from the src dir. the paths in oranda config should always be relative to the oranda config itself and we should look them up that way. we are likely looking them up related to the cwd and that is why they fail when the cwd isn't the same as the path with oranda config

ashleygwilliams:
fixes #95

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -1,14 +1,20 @@
-pub mod analytics;
-mod oranda;
-pub mod project;
-pub mod theme;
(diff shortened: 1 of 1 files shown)


ashleygwilliams:
closing and will sort this using the axoproject integration instead

PR #112, opened 2023-01-27
Revert "fix(css): if path prefix, serve css from prefix"

ashleygwilliams:
Reverts axodotdev/oranda#109


ashleygwilliams:
testing an assumption for progress on #111

PR #128, opened 2023-01-27
Add footer

### LINKED ISSUE #127: footer
was thinking author and license info, maybe another link to the github, similar to our landing page (reuse the design)

SaraVieira:
closes https://github.com/axodotdev/oranda/issues/127

example
CleanShot 2023-01-27 at 20 01 25@2x

fringe pr https://github.com/axodotdev/turbo-axo/pull/195

### DIFF
--- a/tests/mod.rs
+++ b/tests/mod.rs
@@ -183,5 +183,5 @@ fn creates_footer() {
 
     println!("{:?}", site.html);
 
-    assert!(site.html.contains("<footer class=\"axo-gradient flex items-center justify-between px-4 py-2 text-slate-50 text-xs w-full\"><span>2023, My Axo project</span></footer>"));
+    assert!(site.html.contains("<footer class=\"axo-gradient flex items-center justify-between px-4 py-2 text-slate-50 text-xs w-full\"><span>My Axo project</span></footer>"));
 }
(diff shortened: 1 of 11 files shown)


ashleygwilliams (changes requested):
i think we'd need to make the copyright configurable as well- e.g. this says copyright oranda, but it's copyright Axo Developer Co. other projects have even more complicated copyright... e.g. https://github.com/rust-lang/rust/blob/master/COPYRIGHT

SaraVieira:
@ashleygwilliams added tests and removed the copyright

tests/mod.rs
```diff
@@ -175,3 +175,13 @@ fn creates_nav() {
...
 }
+
+#[test]
+fn creates_footer() {
+    let _guard = TEST_RUNTIME.enter();
+    let site = Site::build(&config(), &config().readme_path).unwrap();
+
+    println!("{:?}", site.html);
+
+    assert!(site.html.contains("<footer class=\"axo-gradient flex items-center justify-between px-4 py-2 text-slate-50 text-xs w-full\"><span>2023, My Axo project</span></footer>"));
```
ashleygwilliams: in a year this test will fail haha

ashleygwilliams (approved):
approving, but it's possible we may want to remove the year- at least from the test

tests/mod.rs
SaraVieira: removed the year

PR #130, opened 2023-01-27
More properly read in package.json files, kinda

### LINKED ISSUE #120: get repo from package.json
with #119 i fixed #117 but i am pretty sure deserializing a package.json is gonna have issues. here's how you can specify a repo in package.json: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#repository

Gankra:
~~We can defer to a proper library for the definitions~~ We can write out own definitions and just pull out the parts we want *but* the repository field is just super whack and complex so it's probably not a URL. If you have some npm project that you Definitely Want To Work I can roll up my sleeves and actually iterate on the behaviour.

Fixes #120

### DIFF
--- a/tests/mod.rs
+++ b/tests/mod.rs
@@ -63,6 +63,7 @@ fn it_loads_a_js_project_config() {
             r#"
 {
     "name": "axo",
+    "version": "0.1.0",
     "description": ">o_o<"
 }
     "#,
--- a/tests/mod.rs
+++ b/tests/mod.rs
@@ -64,7 +64,11 @@ fn it_loads_a_js_project_config() {
 {
     "name": "axo",
     "version": "0.1.0",
-    "description": ">o_o<"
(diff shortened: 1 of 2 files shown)


SaraVieira:
Since you are here... Can you please also return the license from it?

SaraVieira:
I know this person wanted to use Oranda
https://github.com/jsx-eslint/eslint-plugin-react/blob/master/package.json

tests/mod.rs
```diff
@@ -63,6 +63,7 @@ fn it_loads_a_js_project_config() {
             r#"
 {
     "name": "axo",
+    "version": "0.1.0",
```
ashleygwilliams, line 66: should we add repo to this as well?

Gankra: done

ashleygwilliams (approved):

PR #131, opened 2023-01-28
Do not error on missing config

### LINKED ISSUE #106: Show warning when no config is found
Right now if no file is there it will error out but it should not, it should only show a warning

SaraVieira:
closes #106


Gankra (approved):
I'm not sure we need a warning but this seems solid

ashleygwilliams (approved):
i might change the warn to an info- since requiring no config is a feature :)

PR #133, opened 2023-01-28
embed markdown themes in final binary

Gankra:


ashleygwilliams (approved):
per convo with @SaraVieira we are gonna merge this and have a few built-ins, but otherwise allow people to provide a path to their theme

PR #135, opened 2023-01-28
refactor: assorted renames and structuring

ashleygwilliams:


src/site/artifacts.rs
```diff
@@ -20,15 +20,15 @@ fn get_kind_string(kind: &ArtifactKind) -> String {
...
+fn build_downloadlink(config: &Config, name: &str) -> String {
     if let (Some(repo), Some(version)) = (&config.repository, &config.version) {
         format!("{}/releases/download/v{}/{}", repo, version, name)
     } else {
         String::new()
     }
 }
 
-pub fn create_artifacts_header(config: &Config) -> Result<Option<Box<div<String>>>> {
+pub fn build_section(config: &Config) -> Result<Option<Box<div<String>>>> {
```
SaraVieira, line 28: Maybe "build_artifacts_section" no?

src/site/layout/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::site::artifacts;
+
+use axohtml::elements;
+use axohtml::{html, text};
+
+pub struct Layout {
+    head: head::Head,
+    theme: String,
+    analytics: Option<Box<elements::script<String>>>,
+    google_script: Option<Box<elements::script<String>>>,
```
SaraVieira, line 18: Do you perfer importing it like this? I find it makes things more confusing

SaraVieira:
I understand that calling things build is a rust thing and I am okay with it but i do think it makes everything more confusing because you end up with 10 things being called build and sometimes without knowing where they are from 

If we do this i do think we need to make the rule of never using just build but make sure to call the module name aswell like header::build

src/site/layout/mod.rs
ashleygwilliams: there are name conflicts between modules and the types exported by axohtml

src/site/artifacts.rs
ashleygwilliams: it's in the artifacts module so i'm trying to not repeat the name, since you'll be calling it `artifacts::build_section`

src/site/layout/mod.rs
SaraVieira: Fair

ashleygwilliams:
so this is a huge PR- and i was pretty stupid to spend so much time futzing on it, but i also think it is useful to think through the structure here. it is probably not smart to land, and esp not now. what i would like is to get folks' perspective on this approach, and then i can make some smaller PRs that move us in this direction. @Gankra @SaraVieira

ashleygwilliams:
closing in favor of a series of smaller PRs that introduce the same changes

PR #136, opened 2023-01-29
Add support for md_book

### LINKED ISSUE #102: support mdbook
in many ways, having an mdbook sidecar is slightly like #46 but not quite. i think we should have something that is configured  like:

```json
"mdbook": "path/to/mdbook/dist/dir"
```

what this enables:
1. add a "Book" link to the nav
2. copy and serve that dir, wholesale, no modifications, at `http://rootURL/path_prefix/book`

thoughts?

SaraVieira:
Question about docs:
- Are we going to use md_book or are we gonna put everything in that one page?

CleanShot 2023-01-29 at 16 43 44

closes https://github.com/axodotdev/oranda/issues/102

### DIFF
--- a/oranda.json
+++ b/oranda.json
@@ -15,5 +15,6 @@
   "artifacts": {
     "cargo_dist": true
   },
-  "homepage": "https://oranda.axo.dev"
+  "homepage": "https://oranda.axo.dev",
+  "md_book": "docs/book"
 }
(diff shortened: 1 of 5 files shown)


ashleygwilliams (approved):
i def think we should use mdbook if we can have the support for it! have you tested this?

SaraVieira:
Yeah i included a gif of it working

PR #141, opened 2023-01-30
Quick docs fix

### LINKED ISSUE #139: inconsistent key name for cargo-dist feature
the current README says to use

```
"artifacts": {
    "cargo-dist": true
}
```

But in fact the key is cargo_dist (presumably because there's no #[serde(rename)] on the field).

Who should change? The code or the docs?

SaraVieira:
Uses `cargo_dist` instead of `cargo-dist`

closes https://github.com/axodotdev/oranda/issues/139

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ using `cargo-dist` you can add this to your `oranda.json`:
 
 ```json
 "artifacts": {
-    "cargo-dist": true
+    "cargo_dist": true
 }
 ```
 


ashleygwilliams (approved):
should we warn when we find "cargo-dist"?

SaraVieira:
We do give an error saying that is not a known name and say what names are possible, right now it says:

✗ >o_o< ERROR: missing field `cargo_dist` at line 17 column 3

Could be better but it does let the person know

PR #147, opened 2023-01-30
Add option for package managers in artifacts

SaraVieira:
related https://github.com/axodotdev/oranda/issues/89

le css https://github.com/axodotdev/turbo-axo/pull/199


src/site/layout/css.rs
```diff
@@ -46,7 +46,7 @@ pub fn fetch_fringe(config: &Config) -> Result<Box<link<String>>> {
     let mut css_file = File::create(css_path)?;
     css_file.write_all(minified_css.as_bytes())?;
 
-    Ok(html!(<link rel="stylesheet" href=css_file_name></link>))
+    Ok(html!(<link rel="stylesheet" href="http://localhost:64679/axo-oranda.css"></link>))
```
SaraVieira: REMEBER TO DELETE

src/site/artifacts/build.rs
```diff
@@ -0,0 +1,67 @@
+use crate::config::Config;
```
ashleygwilliams: i might call this file "page" and then have it have a build command

src/site/artifacts/build.rs
```diff
@@ -0,0 +1,67 @@
...
+use crate::errors::*;
+use crate::site::layout;
+use axohtml::{html, text, unsafe_text};
+use cargo_dist_schema::ArtifactKind;
+
+use super::fetch_manifest;
```
ashleygwilliams: might be good to reorg these imports so that similar things are next to each other- i would also prefer to not use super- maybe that function belongs in this file?

src/site/artifacts/build.rs
```diff
@@ -0,0 +1,67 @@
...
+    }) = &config.artifacts
+    {
+        let mut list = vec![];
+
+        if let Some(Artifacts {
+            cargo_dist: Some(true),
+            ..
+        }) = &config.artifacts
+        {
+            for release in manifest.releases.iter() {
```
ashleygwilliams: it may help with code organization to break some of these loops out into their own functions?

oranda.json
```diff
@@ -13,7 +13,11 @@
     "twitter_account": "@axodotdev"
   },
   "artifacts": {
-    "cargo_dist": true
+    "cargo_dist": true,
+    "package_managers": {
+      "npm": "npm install oranda",
```
ashleygwilliams: do we want this to take a full string or can we construct this from the name of the pkg manager alone?

SaraVieira: I would rather take the string otherwise every time someone makes a new package manager we need to write crappy code

src/site/artifacts/build.rs
SaraVieira: done everywhere.

Also moved the fetch manifest to the cargo dist file

src/site/artifacts/build.rs
SaraVieira: done!

src/site/artifacts/build.rs
SaraVieira: moved some stuff let me know thoughts

oranda.json
ashleygwilliams: maybe this can be a thing we have cargo dist do ;)

src/site/artifacts/build.rs
ashleygwilliams: nice!

ashleygwilliams (approved):
add some docs/tests? and then i think is pretty solid- we are also behind on tests so we can also open an issue for that

SaraVieira:
I added tests since that would be much faster than adding docs for both things, will merge and work on docs

PR #157, opened 2023-02-07
fix assorted paths

### LINKED ISSUE #148: Some things do not take path_prefix into account
- [ ] script tag always loads from root
- [ ] favicon always loads from root
- [ ] link in artifacts to view all is not taking path prefix into account

### LINKED ISSUE #156: downloads should be "install"
we probably want to let people configure this eventually but for the moment i think install makes more sense

SaraVieira:
closes https://github.com/axodotdev/oranda/issues/148

### DIFF
--- a/src/site/layout/header.rs
+++ b/src/site/layout/header.rs
@@ -43,8 +43,7 @@ fn nav(
             .unwrap_or(file_path.as_os_str())
             .to_string_lossy();
 
-        let href =
-            path::generate_prefix_link(path_prefix, format!("{}.html", file_name.to_string()));
+        let href = path::generate_prefix_link(path_prefix, format!("{}.html", file_name));
 
         html.extend(html!(<li><a href=href>{text!(file_name)}</a></li>));
     }
--- a/src/site/layout/header.rs
+++ b/src/site/layout/header.rs
@@ -43,14 +43,14 @@ fn nav(
             .unwrap_or(file_path.as_os_str())
             .to_string_lossy();
 
-        let href = link::generate_link(path_prefix, format!("{}.html", file_name));
+        let href = link::generate(path_prefix, format!("{}.html", file_name));
 
         html.extend(html!(<li><a href=href>{text!(file_name)}</a></li>));
     }
 
     if let Some(artifact) = artifacts {
         if artifact.cargo_dist.is_some() {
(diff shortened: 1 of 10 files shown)


src/site/path.rs
```diff
@@ -0,0 +1,7 @@
+pub fn generate_prefix_link(path_prefix: &Option<String>, file_name: String) -> String {
```
ashleygwilliams: since this does not always generate a link with a prefix should this just be generate? and since the only thing this does is build a link right now.. shall we call it link?

ashleygwilliams (changes requested):

src/site/layout/header.rs
```diff
@@ -42,15 +43,15 @@ fn nav(
...
 
         html.extend(html!(<li><a href=href>{text!(file_name)}</a></li>));
     }
 
     if let Some(artifact) = artifacts {
         if artifact.cargo_dist.is_some() {
-            let href = generate_prefix_link(path_prefix, String::from("artifacts"));
-            html.extend(html!(<li><a href=href>{text!("Downloads")}</a></li>));
+            let href = path::generate_prefix_link(path_prefix, String::from("artifacts.html"));
+            html.extend(html!(<li><a href=href>{text!("Install")}</a></li>));
```
SaraVieira, line 54: closes #156

src/site/path.rs
SaraVieira: hmm? I think you had a train of thought I couldn't follow :/

ashleygwilliams: the function generates the link regardless of if it is prefixed or not, so i think it should be called generate link not generate prefix link.
this file is called path. it has a single function called generate_link so i think it should be called link not path.
the module should be called link, and so the function name should be just generate (or build?)

SaraVieira: OHHH, update the file to be `link.rs` and the fn to be generate_link

src/site/link.rs
```diff
@@ -0,0 +1,7 @@
+pub fn generate_link(path_prefix: &Option<String>, file_name: String) -> String {
```
ashleygwilliams: any reason to keep "link" here?

SaraVieira: fair question, removing

SaraVieira: done

ashleygwilliams (approved):

PR #169, opened 2023-02-08
Better install page

### LINKED ISSUE #163: install methods section of "install" page only shows up if package manager key is present, even tho most install methods displayed are not package managers
Screen Shot 2023-02-07 at 6 13 54 PM

we can show the circled items for just "cargo dist"

SaraVieira:
- Always shows the install section
- Divides script and package manager installs
- gives os's better names

localhost_3000_artifacts (1)

closes https://github.com/axodotdev/oranda/issues/163

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -297,9 +297,9 @@ checksum = "c77df041dc383319cc661b428b6961a005db4d6808d5e12536931b1ca9556055"
 
 [[package]]
 name = "cargo-dist-schema"
-version = "0.0.1"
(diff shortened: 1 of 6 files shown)


ashleygwilliams:
this looks good! i think my only suggestion would be to try to use the "description" in the artifact schema to help give more "instructive text" around the installers (particularly the shell ones), noting from convo with @Gankra in element

ashleygwilliams (approved):
🚀

PR #172, opened 2023-02-09
Only adds  JS to home and artifacts

SaraVieira:
- Removes JS from any additional pages
- Introduces a `needs_js` bool

### DIFF
--- a/src/site/layout/mod.rs
+++ b/src/site/layout/mod.rs
@@ -8,7 +8,7 @@ use crate::config::{analytics, theme, Config};
 use crate::errors::*;
 use axohtml::{html, text, unsafe_text};
 
-pub fn build(config: &Config, content: String) -> Result<String> {
+pub fn build(config: &Config, content: String, need_js: bool) -> Result<String> {
     let theme = theme::css_class(&config.theme);
     let analytics = analytics::get_analytics(config);
     let google_script = match &config.analytics {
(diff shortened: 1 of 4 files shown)


src/site/mod.rs
```diff
@@ -48,9 +48,10 @@ impl Site {
         let dist = &config.dist_dir;
         Self::create_dist_dir(dist)?;
         for page in self.pages {
+            let js = page.needs_js.clone();
```
ashleygwilliams: is this clone necessary?

src/site/layout/mod.rs
```diff
@@ -8,7 +8,7 @@ use crate::config::{analytics, theme, Config};
 use crate::errors::*;
 use axohtml::{html, text, unsafe_text};
 
-pub fn build(config: &Config, content: String) -> Result<String> {
+pub fn build(config: &Config, content: String, need_js: bool) -> Result<String> {
```
ashleygwilliams: we should add needs_js to the Page struct, similar to "is index" and use it from there (also we should standardize on `needs_js`, theres 2 spellings in this PR)

ashleygwilliams (changes requested):

src/site/layout/mod.rs
SaraVieira: my bad, changing everything to needs_js

src/site/mod.rs
SaraVieira: apparently no, removed

ashleygwilliams (approved):
i definitely am gonna swoop in right after this and turn the 2 bools into an options struct for readability sake, but this gets the user facing feature right so approved ;)

SaraVieira:
fixed the tests, waiting for green ci

PR #175, opened 2023-02-09
Fix windows detection

### LINKED ISSUE #167: no recommended download for my OS
It would be nice if the install page helped guide me to the right triple. I see some other issues about some sort of detect OS functionality, but when I check my network tab, that JS isn't being loaded, and I don't see it in the source HTML either.

SaraVieira:
closes https://github.com/axodotdev/oranda/issues/167


steveklabnik (approved):
From my very thorough "run the logic again in my console" testing approach, this looks like it would fix the bug, yep! Nice.

ashleygwilliams (approved):

Issue #144, opened 2023-01-30
allow user to use a local or remote asset as a syntax theme


ashleygwilliams:
currently, we have 2 types of themes: those bundled by syntect and those we "bundle" ourselves - if we continue down this path, users will have to wait for a new release to use a theme we don't provide. instead, we should allow users to also provide a path (either local or remote) to a theme.

ashleygwilliams:
added bug to this because inlining the themes bloated the tool so much it couldn't be published to crates.io

PR #177, opened 2023-02-14
stop inlining themes, remove theme images to reduce binary size

ashleygwilliams:
approaching #144, but not in this PR


ashleygwilliams:
gonna build out a small static file server here to host some themes and docs https://github.com/axodotdev/oranda-asset-server

PR #193, opened 2023-02-27
Add changelog page

### LINKED ISSUE #192: changelogs
currently, like cargo-dist, you can include your `CHANGELOG.md` as an "additional page" for oranda to render, however we can (and should) do better than this, given how important changelogs are for releases.

we should use the github releases api to create a better changelog page for oranda users.

SaraVieira:
closes #192 

fringe pr: https://github.com/axodotdev/turbo-axo/pull/248


Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [package]
 name = "oranda"
 description = "🎁 generate beautiful landing pages for your projects"
-repository = "https://github.com/axodotdev/oranda"
-version = "0.0.1"
```
SaraVieira: UPDATE

src/site/layout/css.rs
```diff
@@ -25,7 +25,7 @@ pub fn build_fringe() -> Box<link<String>> {
     const FRINGE_VERSION: &str = "0.0.10";
     let css_file_name = format!("fringe@{}.css", FRINGE_VERSION);
 
-    html!(<link rel="stylesheet" href=css_file_name></link>)
```
SaraVieira, line 28: UPDATE

oranda.json
```diff
@@ -1,11 +1,11 @@
 {
   "favicon": "https://www.axo.dev/favicon.ico",
-  "path_prefix": "oranda",
```
SaraVieira, line 3: UPDATE

oranda.json
```diff
@@ -1,11 +1,11 @@
...
   "additional_pages": [
     "./docs/configuration.md",
     "./docs/assets.md",
     "./docs/hosting.md"
   ],
+  "changelog": true,
```
ashleygwilliams, line 9: i sorta wonder if we should not make this top level and do something like:

```json
{
...
"pages": {
  "cargo-dist": true,
   "changelog": true,
   }
...
}
```

src/errors.rs
```diff
@@ -13,6 +13,12 @@ pub enum OrandaError {
     #[error(transparent)]
     StripPrefixError(#[from] std::path::StripPrefixError),
 
+    #[error(transparent)]
```
ashleygwilliams: i think when we add errors these days we should try to add full errors w/messages IMO and avoid purely transparent ones unless the error is uniquely good

src/site/changelog/single_release.rs
```diff
@@ -0,0 +1,89 @@
...
+use crate::site::markdown::{self, SyntaxTheme};
+use axohtml::dom::UnsafeTextNode;
+use axohtml::elements::{a, section};
+use axohtml::html;
+use axohtml::{text, unsafe_text};
+use chrono::DateTime;
+
+use super::types::ReleasesApiResponse;
+
+pub fn build_single_release(
```
ashleygwilliams: any reason to name this "single release" and not just "release"

src/site/changelog/single_release.rs
```diff
@@ -0,0 +1,89 @@
...
+        <div class="release-body mb-6">
+            {unsafe_text!(body)}
+        </div>
+        {build_install_button(version, release, path_prefix)}
+    </section>
+    ))
+}
+
+fn build_install_button(
+    version: &Option<String>,
```
ashleygwilliams: version should be required imo

src/site/changelog/single_release.rs
```diff
@@ -0,0 +1,89 @@
...
+use axohtml::html;
+use axohtml::{text, unsafe_text};
+use chrono::DateTime;
+
+use super::types::ReleasesApiResponse;
+
+pub fn build_single_release(
+    release: &ReleasesApiResponse,
+    syntax_theme: &SyntaxTheme,
+    version: &Option<String>,
```
ashleygwilliams: version should be required - and/or what is this version from? we may want to annotate it's origin since there's also a version in the api response right?

src/site/layout/header.rs
```diff
@@ -70,6 +71,15 @@ fn nav(
         html.extend(html!(<li><a href=href>{text!("Docs")}</a></li>));
     };
 
+    if *changelog {
```
ashleygwilliams, line 74: is this `*` required?

src/site/changelog/single_release.rs
SaraVieira: just used to it building react components, changed!

src/site/changelog/single_release.rs
SaraVieira: Version is now required, I show an error if none is present and changed the name of the version param to be config_param, so we know where it comes from

src/site/layout/header.rs
SaraVieira: When not using it I get this error

image

src/errors.rs
SaraVieira: yeah, it's better, I removed this and if there is an error parsing the date I also do not think it makes sense to break everything, I'll just show the original date we have

src/site/changelog/single_release.rs
SaraVieira: fixed!

oranda.json
ashleygwilliams: i'm gonna file an issue for this

PR #196, opened 2023-03-01
Only fetch first non prerelease version

SaraVieira:
related https://github.com/axodotdev/cargo-dist/issues/175

This PR uses the github API instead of using the version in the lock file

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "oranda"
 description = "🎁 generate beautiful landing pages for your projects"
 repository = "https://github.com/axodotdev/cargo-dist"
-version = "0.0.3"
+version = "0.0.4-prerelease.2"
 edition = "2021"
 authors = ["Axo Developer Co. <hello@axo.dev>"]
 license = "MIT OR Apache-2.0"
--- a/src/site/artifacts/cargo_dist.rs
+++ b/src/site/artifacts/cargo_dist.rs
@@ -32,7 +32,7 @@ pub fn fetch_manifest(config: &Config) -> Result<DistManifest> {
             config,
             &String::from("dist-manifest.json"),
             first.tag_name.to_owned(),
-        );
+        )?;
 
         match reqwest::blocking::get(&url)?.error_for_status() {
             Ok(resp) => match resp.json::<DistManifest>() {
@@ -55,7 +55,7 @@ pub fn fetch_manifest(config: &Config) -> Result<DistManifest> {
 }
 
 fn get_installer_path(config: &Config, name: &String, version: String) -> Result<String> {
-    let download_link = create_download_link(config, name, version);
+    let download_link = create_download_link(config, name, version)?;
     let file_string_future = Asset::load_string(download_link.as_str());
     let file_string = tokio::runtime::Handle::current().block_on(file_string_future)?;
     let file_path = format!("{}.txt", &name);
@@ -203,13 +203,13 @@ pub fn build(config: &Config) -> Result<Box<div<String>>> {
     ))
 }
 
-pub fn build_table(manifest: DistManifest, config: &Config) -> Box<div<String>> {
+pub fn build_table(manifest: DistManifest, config: &Config) -> Result<Box<div<String>>> {
     let mut table = vec![];
     for release in manifest.releases.iter() {
         for artifact_id in release.artifacts.iter() {
             let artifact = &manifest.artifacts[artifact_id];
             if let Some(name) = artifact.name.clone() {
-                let url = create_download_link(config, &name, release.app_version.to_owned());
+                let url = create_download_link(config, &name, release.app_version.to_owned())?;
                 let kind = get_kind_string(&artifact.kind);
                 let targets: &String = &artifact.target_triples.clone().into_iter().collect();
                 table.extend(vec![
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "oranda"
 description = "🎁 generate beautiful landing pages for your projects"
 repository = "https://github.com/axodotdev/cargo-dist"
-version = "0.0.3"
+version = "0.0.4-prerelease.2"
```
ashleygwilliams, line 5: we should bring this back to oranda's version number (0.0.1)

src/site/artifacts/cargo_dist.rs
```diff
@@ -21,25 +22,40 @@ pub fn get_os(name: &str) -> Option<&str> {
...
+            config,
+            &String::from("dist-manifest.json"),
+            Some(first.tag_name.to_owned()),
+        );
 
-    match reqwest::blocking::get(&url)?.error_for_status() {
-        Ok(resp) => match resp.json::<DistManifest>() {
-            Ok(manifest) => Ok(manifest),
-            Err(e) => Err(OrandaError::CargoDistManifestParseError {
+        match reqwest::blocking::get(&url)?.error_for_status() {
```
ashleygwilliams, line 37: we should add a blocking api to axo asset because we really should be using it for these fetches because of the error handling

ashleygwilliams: (i am the one who added the errors heh hey but yeah)

src/site/artifacts/cargo_dist.rs
```diff
@@ -21,25 +22,40 @@ pub fn get_os(name: &str) -> Option<&str> {
...
+        Err(OrandaError::Other(
+            "Repository is mandatory for the cargo dist option".to_owned(),
+        ))
     }
 }
 
-fn get_installer_path(config: &Config, name: &String) -> Result<String> {
-    let download_link = create_download_link(config, name);
+fn get_installer_path(config: &Config, name: &String, version: String) -> Result<String> {
+    let download_link = create_download_link(config, name, Some(version));
```
ashleygwilliams: download link take an optional version? hrmmm

src/site/artifacts/cargo_dist.rs
```diff
@@ -292,3 +291,16 @@ pub fn build_list(manifest: &DistManifest, config: &Config) -> Result<Box<div<St
     </div>
     ))
 }
+
+fn create_download_link(config: &Config, name: &String, version: Option<String>) -> String {
```
ashleygwilliams: version should just be a string IMO- unless there's a case where it is actually optional? (everything i've read in this diff so far manually adds `Some`)

src/site/artifacts/cargo_dist.rs
```diff
@@ -292,3 +291,16 @@ pub fn build_list(manifest: &DistManifest, config: &Config) -> Result<Box<div<St
...
+    if let (Some(repo), Some(v)) = (&config.repository, version) {
+        let version_to_use = if v.contains('v') {
+            v.split('v').collect::<Vec<&str>>()[1]
+        } else {
+            v.as_str()
+        };
+        format!("{}/releases/download/v{}/{}", repo, version_to_use, name)
+    } else {
+        String::new()
```
ashleygwilliams: if there's no repo or version i feel like we should error here?

src/site/changelog/mod.rs
```diff
@@ -32,61 +32,65 @@ fn build_prerelease_toggle(releases: Vec<ReleasesApiResponse>) -> Option<Box<div
...
-        let mut releases_nav: Vec<Box<li<String>>> = vec![];
         let url = format!(
             "https://api.github.com/repos/{}/{}/releases",
             url_parts[0], url_parts[1]
         );
 
         let releases = reqwest::blocking::Client::new()
-            .get(&url)
+            .get(url)
             .header(USER_AGENT, "oranda")
```
ashleygwilliams: good call, we may also want to add version number :)

Cargo.toml
SaraVieira: In the other PR since I still need the latest version of cargo dist, this merges to changelog PR

src/site/artifacts/cargo_dist.rs
SaraVieira: good catch

src/site/artifacts/cargo_dist.rs
SaraVieira: you are right, fixed!

src/site/artifacts/cargo_dist.rs
SaraVieira: Done!

src/site/changelog/mod.rs
SaraVieira: How can I get that?

SaraVieira: found it and done

src/site/artifacts/cargo_dist.rs
```diff
@@ -5,6 +5,7 @@ use cargo_dist_schema::{Artifact, ArtifactKind, DistManifest, Release};
 
 use crate::config::Config;
 use crate::errors::*;
+use crate::site::changelog::fetch_releases;
```
ashleygwilliams: i think we should import changelog, not the function name

SaraVieira: updated all the imports there

PR #200, opened 2023-03-07
move css over to oranda

SaraVieira:
This PR moves all the CSS we use to the oranda repo.

* It creates a new folder that generates this CSS 
* deletes unused things  in fringe when copying over
* moves most classes in the HTML to CSS for easier theming


docs/configuration/theme.md
```diff
@@ -19,9 +19,7 @@ Dark is the default theme.
 
 ## Customizing Themes
 
-Themes can be further customized by adding extra CSS. 
-
-To see the styles applied by the theme, you can view [Fringe's `theme` directory on `unpkg`](https://www.unpkg.com/browse/@axodotdev/fringe/themes/).
+Themes can be further customized by adding extra CSS.
```
ashleygwilliams, line 22: this feels repetitive to the next line

ashleygwilliams (changes requested):
in general i think this is looking good- i do think we should not nest the css in the rust source code.. we should pull it up to the top as an additional project almost. it can't be treated like a local asset because it will not be bundled in the binary- so we will need to write an action to publish the css to github releases- where we can fetch it

SaraVieira:
About your comment, i assumed that was the idea, i would make it work this way, copy everything and then you would come to the PR and do your releasing magic so we could merge

src/site/layout/footer.rs
```diff
@@ -23,7 +23,7 @@ pub fn create_footer(config: &Config) -> Box<footer<String>> {
     );
 
     html!(
-        <footer class="axo-gradient text-slate-50 flex w-full justify-between px-4 py-2 text-xs items-center">
+        <footer class="footer">
```
ashleygwilliams, line 26: should this be a class or should it just target the footer element?

ashleygwilliams: @SaraVieira ^

ashleygwilliams: filed as #205

ashleygwilliams (approved):
still has issues (#212, #204) but let's merge this

PR #201, opened 2023-03-08
skip & warn user when additional page is not md

### LINKED ISSUE #151: we should warn or error if an additional file is not markdown

SaraVieira:
closes https://github.com/axodotdev/oranda/issues/151


src/site/layout/header.rs
```diff
@@ -3,7 +3,7 @@ use std::path::Path;
 use crate::config::artifacts::Artifacts;
 use crate::config::Config;
 use crate::errors::*;
-use crate::site::link;
+use crate::site::{link, Site};
```
ashleygwilliams: is there a reason you put the function in Site? i would probably put it in page...

src/site/mod.rs
```diff
@@ -1,3 +1,4 @@
+use std::ffi::OsStr;
```
ashleygwilliams: ... i'm not sure we need this. are we using ffi?

src/site/mod.rs
```diff
@@ -24,8 +26,16 @@ impl Site {
...
         if let Some(files) = &config.additional_pages {
             for file in files {
-                let additional_page = Page::new_from_file(config, file, false)?;
-                pages.push(additional_page)
+                if Self::is_file_markdown(file) {
+                    let additional_page = Page::new_from_file(config, file, false)?;
+                    pages.push(additional_page)
+                } else {
+                    let msg = format!(
+                        "File {} in additional pages is not markdown and will be skipped",
```
ashleygwilliams, line 34: nice

ashleygwilliams (changes requested):
structurally, checking whether a file is markdown is checking the "source" of a page- i would probably org this such that theres page/source.rs so you could say Page::source.is_markdown()

src/site/mod.rs
SaraVieira: I got this from the example here: https://doc.rust-lang.org/std/path/struct.Path.html

Not sure what else to use to make it a str

src/site/layout/header.rs
SaraVieira: Honestly forgot about page.

Done some moving, let me know if it looks good

src/site/mod.rs
```diff
@@ -74,4 +84,11 @@ impl Site {
 
         Ok(())
     }
+
+    pub fn is_file_markdown(file: &str) -> bool {
+        let file_path = Path::new(&file);
+        let extension = file_path.extension().and_then(OsStr::to_str);
```
ashleygwilliams:
```suggestion
        let extension = match file_path.extension().to_str() {
            Some(ext) => ext,
            None => OrandaError::NonAsciiFileExtension { file_path },
        };
```

src/site/layout/header.rs
```diff
@@ -42,15 +42,17 @@ fn nav(
...
-            let file_path = Path::new(page);
-            let file_name = file_path
-                .file_stem()
-                .unwrap_or(file_path.as_os_str())
-                .to_string_lossy();
+            if page::source::is_markdown(page) {
+                let file_path = Path::new(page);
+                let file_name = file_path
+                    .file_stem()
+                    .unwrap_or(file_path.as_os_str())
```
ashleygwilliams: can you say more about the logic here?

SaraVieira: updated with a warning

ashleygwilliams (approved):

PR #208, opened 2023-03-09
Create new themes

SaraVieira:
- Adds new hacker theme
- Adds new cupcake theme
- Moves files arround to make more sense in our current setup
- Removes layers as they were causing issues and we want things to be global


oranda-css/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "tailwind.config.js",
   "scripts": {
     "build": "tailwindcss -c tailwind.config.js -i ./css/main.css -o ./dist/oranda.css --minify",
-    "dev": "npm run build --watch"
```
SaraVieira, line 8: hmmm, this wasn't working

oranda.json
```diff
@@ -1,4 +1,5 @@
 {
+  "theme": "hacker",
```
SaraVieira: remove after testing

oranda-css/css/base.css
```diff
@@ -1,94 +1,173 @@
 @tailwind base;
 
-@layer base {
```
ashleygwilliams, line 3: curious why we don't want the base to be a layer?

oranda-css/css/themes/cupcake.css
```diff
@@ -0,0 +1,96 @@
+@import url("https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600&display=swap");
+
+html.cupcake body {
```
ashleygwilliams, line 3: building themes like this feels burdensome - having to add to class to html for every selector... can you share more about this design and the possible alternatives?

SaraVieira: The main alternative that I see is we move to that supports nesting but that would mean we would mean we would need to use the POSTCSS cli instead of the tailwind one

oranda-css/css/base.css
SaraVieira: I think because of how layers and content in tailwind was working I wasn't getting some of the styles

SaraVieira:
Yaaaaaaaasssss

PR #218, opened 2023-03-14
Ensure we delete the dist directory before building

### LINKED ISSUE #210: delete public directory before building

SaraVieira:
- Removes scattered calls to function that ensured there was a dist dir
- Adds one function on the build command to create the dir or if it exists to delete it and create it

closes https://github.com/axodotdev/oranda/issues/210

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -25,6 +25,11 @@ pub enum OrandaError {
     #[error(transparent)]
     Reqwest(#[from] reqwest::Error),
 
+    #[error(
+        "Failed to create a directory, `{dist_path}` to build your project in. Details:\n{details}"
+    )]
+    DistDirCreationError { dist_path: String, details: String },
+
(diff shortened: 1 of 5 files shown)


src/commands/build.rs
```diff
@@ -28,4 +29,15 @@ impl Build {
         Message::new(MessageType::Success, &msg).print();
         Ok(())
     }
+
+    pub fn ensure_clean_dist_dir(dist_path: &String) -> Result<()> {
```
ashleygwilliams: this logic has a fair amount of redundancy, i think you could write it:

```
if the path exists {
   delete;
}
create
```

ashleygwilliams (changes requested):

src/commands/build.rs
ashleygwilliams: `&String`  should be `&str`

ashleygwilliams (approved):

PR #219, opened 2023-03-14
Use announcement_changelog in changelog body if dist-manifest is present

SaraVieira:
Uses the announcement_changelog in case the release we are showing has a dist manifest

related: https://github.com/axodotdev/cargo-dist/issues/199

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,6 +69,55 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e579a7752471abc2a8268df8b20005e3eadd975f585398f17efcfd8d4927371"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "is-terminal",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


src/site/changelog/github_release.rs
```diff
@@ -97,7 +86,24 @@ impl GithubRelease {
...
+                    let manifest = cargo_dist::fetch_manifest(config)?;
+                    match manifest.announcement_changelog {
+                        Some(changelog) => changelog,
+                        None => md.to_string(),
+                    }
+                } else {
+                    md.to_string()
+                }
+            }
+            None => String::new(),
```
ashleygwilliams: can you say more about this condition- this doesn't seem correct to me...

src/site/changelog/github_release.rs
```diff
@@ -97,7 +86,24 @@ impl GithubRelease {
...
+        let body_md = match &self.body {
+            Some(md) => {
+                if self.has_dist_manifest() {
+                    let manifest = cargo_dist::fetch_manifest(config)?;
+                    match manifest.announcement_changelog {
+                        Some(changelog) => changelog,
+                        None => md.to_string(),
+                    }
+                } else {
+                    md.to_string()
```
ashleygwilliams: do we need a String or would a &str work

src/site/changelog/github_release.rs
SaraVieira: If there is no body I just pass an empty string since the body is actually optional in cargo dist

src/site/changelog/github_release.rs
SaraVieira: let me change this to str

src/site/changelog/github_release.rs
ashleygwilliams: in that case we should be using an option i think?

PR #220, opened 2023-03-15
move toggle and fix spacing

### LINKED ISSUE #202: changelog page css
Screen Shot 2023-03-08 at 8 49 20 AM

can we balance the spacing between the releases heading, the header and the body of the release page? there's more space below than above and it feels odd. i also think we could potentially move the toggle to the "sidebar" that it is toggling, instead of having it above?

SaraVieira:
- Moves toggle to the nav
- Makes spacing more consistent

image

closes https://github.com/axodotdev/oranda/issues/202

### DIFF
--- a/oranda-css/css/pages/changelog.css
+++ b/oranda-css/css/pages/changelog.css
@@ -20,7 +20,7 @@
 }
 
 .releases-nav {
-  @apply top-12 sticky self-start;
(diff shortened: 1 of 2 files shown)


ashleygwilliams (approved):
should i make a new oranda-css release?

SaraVieira:
Honestly up to you there, I would say its not needed and we do it next week

Issue #236, opened 2023-03-28
use syntax highlight theme that supports `toml`


spitfire05:
## Oranda version

Built from `544e447caf0ecc7d1c7f2d685842755f2538023e` (today's HEAD).

## What is wrong

Using [portproxy](https://github.com/spitfire05/portproxy) to test.

In the README.md file I have a `toml`-annotated code section, that is not being rendered to oranda's html:

```html
<p>Config should contain one or more <code>[[proxy]]</code> elements, that define the port mappings:</p>
<!-- Code block should be here! -->

<h2>Run</h2>
```

However, if I remove the `toml` annotation from markdown, the code block is properly rendered (even with some basic highlighting, it seems).

```html
<p>Config should contain one or more <code>[[proxy]]</code> elements, that define the port mappings:</p>
<pre style="background-color:#263238;"><span style="color:#89ddff;">[[</span><span style="color:#eeffff;">proxy</span><span style="color:#89ddff;">]]
</span><span style="color:#eeffff;">listen </span><span style="color:#89ddff;">= "</span><span style="color:#c3e88d;">localhost:8080</span><span style="color:#89ddff;">"        #</span><span style="color:#eeffff;"> local address to listen on
</span><span style="color:#eeffff;">connect </span><span style="color:#89ddff;">= "</span><span style="color:#c3e88d;">some-server.lan:8485</span><span style="color:#89ddff;">" #</span><span style="color:#eeffff;"> remote </span><span style="color:#89ddff;">(</span><span style="color:#eeffff;">or local</span><span style="color:#89ddff;">)</span><span style="color:#eeffff;"> address to connect to
</span></pre>

<h2>Run</h2>
```

## Expected behavior

Markdown code blocks annotated with `toml` should be properly rendered to output html.

ashleygwilliams:
hey @spitfire05 ! thanks for filing! on a naive glance, i *suspect* this is related to the syntax highlighting theme/library not supporting toml and therefore returning None somewhere in the build, causing the block to just be dropped entirely as opposed to rendered but not highlighted! will take a look in a bit cc @SaraVieira

ashleygwilliams:
by the way- we started a discord if you'd like to join us and chat there!! would love to have you: https://discord.gg/A8QQkCM5

spitfire05:
[...]

PR #237, opened 2023-03-29
Default to text when no syntax is found

SaraVieira:
When no syntax is found in the code block passed we used to give an error, this changes that and makes the default to be the `plain text` one is used 

related https://github.com/axodotdev/oranda/issues/236

### DIFF
--- a/src/site/markdown/syntax_highlight/mod.rs
+++ b/src/site/markdown/syntax_highlight/mod.rs
@@ -24,9 +24,15 @@ fn find_syntax<'a>(ps: &'a SyntaxSet, language: &'a str) -> Result<&'a SyntaxRef
         return Ok(syntax_name);
     }
 
-    Err(OrandaError::Other(
-        "Please add the language to your code snippets".to_owned(),
-    ))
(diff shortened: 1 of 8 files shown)


src/site/markdown/syntax_highlight/mod.rs
```diff
@@ -24,9 +25,18 @@ fn find_syntax<'a>(ps: &'a SyntaxSet, language: &'a str) -> Result<&'a SyntaxRef
...
-    ))
+    // this syntax will always be found as it's part of the default set
+    // https://github.com/sublimehq/Packages
+    let plain_text = ps
+        .syntaxes()
+        .iter()
+        .find(|syntax| syntax.name == "Plain Text")
+        .unwrap();
+
+    let msg = format!("Found syntax highlight language annotation `{language}` which is not currently supported. The annotated block will be shown as plaintext. Please file an issue https://github.com/axodotdev/oranda/issues/new to let us know you'd like to see it supported.");
```
SaraVieira: Maybe only show this if the user passed a language? If no language is passed then we are supposed to show it as plain text

PR #238, opened 2023-03-29
WIP: migrate to miette

Gankra:
currently this produces output like:

image

(I imagine you'll want that to be better...)

This PR has three fundamental parts (all done together since they have to be done in lockstep):

* update to the miette version of axoasset (currently via git, so super don't merge)
* update OrandaError to be a proper miette::Diagnostic
* do a crapton of stuff in main to register a bunch of hooks and mess with colors and json output

Unlike libraries, applications "should" register hooks to produce final miette output (otherwise they'll behave like normal thiserror Errors). Also note that to emit a miette Diagnostic you need to:

* wrap it in a miette::Report
* *Debug* format it (`{:?}`)

This is for... Reasons that I can't remember the full details of. If you Display format it (`{}`) you'll get thiserror-style output. I forget what the Report does, if anything. Honestly I might be cargo-culting that at this point.

json error output is currently dead code, but I figured I'd toss it in there tentatively since you'll probably want it for Web Services. A lot of this code is messy gunk that's been copied several times. This version is copied from cargo-dist. If there's anything you Hate there's probably a decent argument for it. I think for instance the current panic hook will swallow the actual backtrace..? I thought I fixed this in cargo-dist but I might be wrong. I guess technically it's prettier but... bleh.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,30 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a76fd60b23679b7d19bd066031410fb7e458ccc5e958eb5c325888ce4baedc97"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
+name = "ahash"
+version = "0.7.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


Gankra:
I've never bothered to configure it before but there's a bunch of stuff in miette to tune the fancy output that we could probably use to make things more seamless: 

* https://docs.rs/miette/latest/miette/struct.GraphicalTheme.html
* https://docs.rs/miette/latest/miette/struct.GraphicalReportHandler.html

Gankra:
Oh yeah, and if you want to wrap another Diagnostic, you need to add `#[diagnostic(transparent)]` in addition to `#[error(transparent)]` to forward everything.

For instance adding that to the error that just forwards AxoAsset issues changes the screenshot in the first comment to this one:

image

(note that "help" properly funneled through)

ashleygwilliams:
i'm likely gonna integrate the work here into https://github.com/axodotdev/oranda/pull/241 and close it- there's not a super clean way for me to pull commits out of this unfortunately

ashleygwilliams:
https://github.com/axodotdev/oranda/pull/241/commits/574bd8033a5055152ff693570dc492cf2b1f3297

PR #240, opened 2023-03-30
dev mode

### LINKED ISSUE #209: oranda dev command
basically oranda build --watch + oranda serve

ashleygwilliams:
fixes #209 

- [x] add dev command
- [ ] add watch


src/commands/dev.rs
```diff
@@ -0,0 +1,27 @@
+use std::path::PathBuf;
+
+use clap::Parser;
+
+use crate::commands::{Build, Serve};
+use oranda::errors::*;
+
+#[derive(Debug, Parser)]
+pub struct Dev {
+    #[arg(long, default_value = "7979")]
```
emberian: nice

src/commands/serve.rs
```diff
@@ -17,6 +17,12 @@ pub struct Serve {
 }
 
 impl Serve {
+    pub fn new(port: Option<u16>) -> Self {
+        Serve {
+            port: port.unwrap_or(7979),
```
emberian:
```suggestion
            port: port.or_else(Dev::default().port)
```

if this worked (will the derive pick up the clap default?), this is still a little noisy to me compared to 7979 but might be nice? wdyt?

ashleygwilliams: if this worked i would 100% prefer it heh, let's see

ashleygwilliams:
gonna add watch in the next version, this is fine for now

PR #241, opened 2023-04-04
integrate axocli

ashleygwilliams:


src/main.rs
```diff
@@ -1,30 +1,42 @@
...
 
 #[derive(Subcommand, Debug)]
 enum Command {
     Build(Build),
     Serve(Serve),
 }
+
 fn main() {
     let runtime = tokio::runtime::Builder::new_multi_thread()
         .worker_threads(1)
```
Gankra: it *probably* won't matter in this case but ideally this kind of thing is setup within `run` so that it's all properly shutdown when `run` returns/panics. `start` will call `std::process::exit` on error, so any dtors that are scoped to main will never run.

Gankra: (Cli::parse() is fine/intended to be in main tho)

src/message.rs
```diff
@@ -64,3 +65,12 @@ impl Message {
         }
     }
 }
+
+/// Style of output we should produce
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, ValueEnum)]
+pub enum OutputFormat {
```
Gankra, line 71: A pedantic detail in case relevant: OutputFormat derives ValueEnum which is a clap-derive thing. As such, message.rs now depends on clap. That could be a problem if you had dreams of a really hard library/binary split but I don't think you do, so this is fine!

Gankra (approved):
looks good! some minor comments

src/message.rs
ashleygwilliams: yeah i hear ya, makes sense. if/when we end up wanting that, it's an easy enough change

src/main.rs
ashleygwilliams: ah word, i mostly just ignored the runtime thing because i know the goal is to eventually get that into axocli- but i agree with your point and will move it before merge

PR #274, opened 2023-04-19
refactor: separating presentation from data, at least a little

### LINKED ISSUE #226: refactor artifacts/changelog code

ashleygwilliams:
fixes #226


src/data/cargo_dist.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub source: GithubRelease,
+}
+
+pub fn get_os(name: &str) -> Option<&str> {
+    match name.trim() {
+        "x86_64-unknown-linux-gnu" => Some("linux"),
+        "x86_64-apple-darwin" => Some("mac"),
+        "aarch64-apple-darwin" => Some("arm mac"),
+        "x86_64-pc-windows-msvc" => Some("windows"),
+        &_ => None,
```
Gankra, line 22: note to self: more good stuff to maybe put in dist-schema as an api (i know this is just being moved and already existed)

src/data/github/mod.rs
```diff
@@ -0,0 +1,88 @@
+use crate::errors::*;
+
+use reqwest::header::USER_AGENT;
+use serde::{Deserialize, Serialize};
+
+mod repo;
+pub use repo::GithubRepo;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct GithubRelease {
```
Gankra, line 12: transcribing github's docs is a pain but it might be good to link any relevant docs here?

src/data/github/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub updated_at: String,
+    pub browser_download_url: String,
+}
+
+impl GithubRelease {
+    pub fn fetch_all(repo: &GithubRepo) -> Result<Vec<GithubRelease>> {
+        let releases_url = format!(
+            "https://octolotl.axodotdev.host/releases/{}/{}",
+            repo.owner, repo.name
+        );
```
Gankra: well that's one way to deal with rate-limits... should we have a fallback to hit github directly if this is down?

src/data/github/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    pub fn has_dist_manifest(&self) -> bool {
+        for asset in &self.assets {
+            if asset.name == "dist-manifest.json" {
+                return true;
+            }
+        }
+        false
```
Gankra:
```suggestion
        self.assets.iter().any(|a| a.name == "dist-manifest.json")
```

nice place for `any`

src/data/github/repo.rs
```diff
@@ -0,0 +1,36 @@
+use crate::errors::*;
+
+use miette::{miette, IntoDiagnostic};
+use url::Url;
+
+#[derive(Debug)]
```
Gankra:
```suggestion
#[derive(Debug, Clone)]
```

I always kick myself later when i leave this off but obviously not important

src/data/github/repo.rs
```diff
@@ -0,0 +1,36 @@
...
+pub struct GithubRepo {
+    pub owner: String,
+    pub name: String,
+}
+
+impl GithubRepo {
+    pub fn from(repo_url: &str) -> Result<Self> {
```
Gankra: kinda unconventional to name a Not An Impl Of The From Trait function "from". Usually you'd call it `from_repo_url` or something.

src/data/github/repo.rs
```diff
@@ -0,0 +1,36 @@
...
+        let repo_parsed = match Url::parse(repo_url).into_diagnostic() {
+            Ok(parsed) => Ok(parsed),
+            Err(e) => Err(OrandaError::RepoParseError {
+                repo: repo_url.to_string(),
+                details: e,
+            }),
+        };
+        let binding = repo_parsed?;
```
Gankra:
```suggestion
        let binding = Url::parse(repo_url).into_diagnostic().map_err(|e| OrandaError::RepoParseError {
            repo: repo_url.to_string(),
            details: e,
        })?;
```

nice place for map_err

src/data/mod.rs
```diff
@@ -1,3 +1,55 @@
...
-pub mod changelog;
-pub mod releases;
+use crate::data::github::{GithubRelease, GithubRepo};
+use crate::errors::*;
+
+pub mod cargo_dist;
+use cargo_dist::DistRelease;
+pub mod github;
+mod release;
+pub use release::Release;
```
Gankra, line 9: nit: unconventional interspersing of mod and use

src/data/mod.rs
```diff
@@ -1,3 +1,55 @@
...
+        let mut all = vec![];
+        for gh_release in gh_releases {
+            if gh_release.prerelease {
+                has_prereleases = true
+            }
+            if !found_latest_dist_release && gh_release.has_dist_manifest() {
+                let release = Release::new(gh_release.clone())?;
+                if let Some(manifest) = release.manifest {
+                    found_latest_dist_release = true;
+                    latest_dist_release = Some(DistRelease {
```
Gankra:
```suggestion
        let mut latest_dist_release = None;
        let mut all = vec![];
        for gh_release in gh_releases {
            if gh_release.prerelease {
                has_prereleases = true
            }
            if latest_dist_release.is_none() && gh_release.has_dist_manifest() {
                let release = Release::new(gh_release.clone())?;
                if let Some(manifest) = release.manifest {
                    latest_dist_release = Some(DistRelease {
```

found_latest_dist_release seems redundant

Gankra: also are we definitely iterating the releases in reverse chronological order? otherwise this will get the first one, no? (worth a comment in the code)

src/errors.rs
```diff
@@ -50,8 +50,11 @@ pub enum OrandaError {
         details: reqwest::Error,
     },
 
-    #[error("Could not find any releases from {repo} with a cargo-dist compatible `dist-manifest.json`.")]
-    NoCargoDistReleasesFound { repo: String },
+    #[error("Could not find any releases from {repo_owner}/{repo_name} with a cargo-dist compatible `dist-manifest.json`.")]
```
Gankra, line 53: backticks not terribly useful here..? unless maybe it helps sunfish render..?

src/data/cargo_dist.rs
ashleygwilliams: this is definitely one of those functions i suspect i might move every refactor because i can't decide where it goes lol, def would be great to put in the cargo dist util

src/site/artifacts/installers.rs
```diff
@@ -0,0 +1,211 @@
...
+        if let (Some(install_hint), Some(name)) = (&current_hint.install_hint, &current_hint.name) {
+            let file_path =
+                cargo_dist::write_installer_source(config, name, &data.app.app_version)?;
+            Ok((install_hint.to_string(), file_path))
+        } else {
+            Err(no_hint_error)
+        }
+    } else {
+        Err(no_hint_error)
+    }
```
Gankra: interesting approach, i don't hate it

tests/build/mod.rs
```diff
@@ -1,168 +1,237 @@
...
 use fixtures::{oranda_config, page};
 
+fn temp_build_dir() -> String {
+    assert_fs::TempDir::new()
+        .unwrap()
+        .to_str()
+        .unwrap()
+        .to_string()
+}
+
```
Gankra:
```suggestion
fn temp_build_dir() -> (TempDir, String) {
    let dir = assert_fs::TempDir::new().unwrap();
    (dir, dir.to_str().unwrap().to_string())
}
```

Yeah this won't work. The dir gets cleaned up by TempDir's Drop. You need to return it.

Caller would then do like

```rs
let (_guard, temp_dir) = temp_build_dir();
```

Alternatively only return the TempDir and make the user do .to_str() themselves

src/site/artifacts/header.rs
Gankra, line 1: empty file?

src/site/layout/javascript/mod.rs
```diff
@@ -1,15 +1,40 @@
...
+        let snippet = analytics::snippet(config);
+        match &config.analytics {
+            Some(analytics::Analytics::Google(g)) => {
+                let google_script = Some(g.get_script());
+                Ok(Self {
+                    snippet,
+                    google_script,
+                })
+            }
+            _ => Ok(Self {
```
Gankra, line 26: nice save of three chars to type None

(just kidding around)

src/site/layout/mod.rs
```diff
@@ -1,85 +1,88 @@
...
+        let meta_tags = head::create_meta_tags(config);
+        let favicon = if let Some(favicon) = config.favicon.clone() {
+            Some(head::get_favicon(
+                favicon,
+                config.dist_dir.clone(),
+                &config.path_prefix,
+            )?)
+        } else {
+            None
+        };
```
Gankra, line 50: use `map`

Gankra (approved):
this all seems reasonable, left some nits

found the issue you had with tests

src/data/github/mod.rs
ashleygwilliams: added!

src/data/github/mod.rs
ashleygwilliams: added!

src/data/github/repo.rs
ashleygwilliams: yeeeeah this was a panic name lol, i'll update

src/site/artifacts/installers.rs
ashleygwilliams: heh, i updated it

src/data/mod.rs
ashleygwilliams: so yes they are returned in chronological order- but i did wonder about this - i can't find any documentation that guarantees this tho?

src/errors.rs
ashleygwilliams: yeah i want to keep it for sunfish rendering

src/site/artifacts/header.rs
ashleygwilliams: removed!

PR #275, opened 2023-04-20
fix contrast in repo banner

### LINKED ISSUE #270: GitHub header contrast is too low
Untitled

None of the colors in the gradient work too well with the link hover color, this should definitely just stay black, white or another color with high contrast.

SaraVieira:
closes https://github.com/axodotdev/oranda/issues/270

### DIFF
--- a/oranda-css/css/components.css
+++ b/oranda-css/css/components.css
@@ -25,5 +25,7 @@ footer {
 }
 
 .repo_banner > a {
-  @apply flex justify-center gap-2 items-start text-slate-50 dark:text-axo-black h-[20px];
+  @apply flex justify-center gap-2 items-start 
+  hover:text-slate-50
+  text-slate-50 dark:text-axo-black dark:hover:text-axo-black h-[20px];
 }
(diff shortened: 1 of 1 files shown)


oranda-css/css/components.css
```diff
@@ -25,5 +25,5 @@ footer {
 }
 
 .repo_banner > a {
-  @apply flex justify-center gap-2 items-start text-slate-50 dark:text-axo-black h-[20px];
+  @apply flex justify-center gap-2 items-start hover:text-slate-50 text-slate-50 dark:text-axo-black dark:hover:text-axo-black h-[20px];
```
ashleygwilliams: lol i hate how git shows this diff hahaha

what's the behavior here? i do think on hover it should do something- maybe an underline? i can't tell exactly what this does but my sense is that it just keeps the text black, which is better than it disappearing, but probably not all we want to happen

ashleygwilliams (commented):
can you add a video or screencap? i think this is likely to fix the problem of the disappearing text, but not give the user on feedback that we'd like

SaraVieira:
@ashleygwilliams updated

CleanShot 2023-04-20 at 14 32 06

ashleygwilliams (approved):

PR #284, opened 2023-05-04
Build changelogs in separate pages

manyinsects:
With this PR, every changelog item now gets rendered at `changelog/{tagname}.html`.

CleanShot 2023-05-04 at 13 47 06@2x

I've left the existing index page largely unchanged, with the exception of making the heading and sidebar links link to the respective changelog page, but we might want to consider shortening the release body for brevity or something.

Current concerns:

- I've had to modify the CSS link to become absolute, it should work with sites with path prefixes, but what about sites without? Need to test this
- Are all tag names URL-safe (I assume not) or do we need to run them through a slug function?
- Do we want a link back to the changelog index page on every individual page?

### DIFF
--- a/src/site/changelog.rs
+++ b/src/site/changelog.rs
@@ -21,7 +21,7 @@ pub fn build(context: &Context, config: &Config) -> Result<String> {
 
         let link = format!("#{}", &release.source.tag_name);
 
-        releases_html.extend(build_page_preview(release, config)?);
+        releases_html.extend(build_page_preview(release, config, true)?);
         releases_nav.extend(
             html!(<li class=classnames><a href=link>{text!(&release.source.tag_name)}</a></li>),
         )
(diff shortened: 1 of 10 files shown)


src/site/changelog.rs
```diff
@@ -44,7 +44,50 @@ pub fn build(context: &Context, config: &Config) -> Result<String> {
...
+            build_single_release(context, config, release)?,
+        ))
+    }
+
+    Ok(releases)
+}
+
+/// Builds a single, standalone release page.
+pub fn build_single_release(
+    _context: &Context,
```
ashleygwilliams: is there a reason to pass context if we don't need it?

src/site/layout/css.rs
```diff
@@ -24,7 +25,8 @@ fn concat_minify(css_files: &[String]) -> Result<String> {
     Ok(css)
 }
 
-pub fn build_oranda(dist_dir: &str) -> Result<Box<link<String>>> {
+pub fn build_oranda(config: &Config) -> Result<Box<link<String>>> {
```
ashleygwilliams: i might just pass dist_dir and path_prefix here so it's clear what this function needs

ashleygwilliams (commented):
leaving some comments, but i'm about to pull this down and update :) am excited to get this merged

ashleygwilliams (approved):

src/site/changelog.rs
manyinsects: nope, _maybe_ in the future but for now since we have one caller we can remove that parameter

src/site/layout/css.rs
manyinsects: works for me

Issue #190, opened 2023-02-25
re add mdbook now that we support it


ashleygwilliams:

ashleygwilliams:
closed by #285

PR #285, opened 2023-05-04
setup minimal book from docs

Gankra:
fixe #190


.github/workflows/web.yml
```diff
@@ -1,32 +1,101 @@
...
+      - completed
 
+# Alright, let's do it!
 jobs:
   web:
     name: Build and deploy site and docs
     runs-on: ubuntu-latest
     steps:
+      # Setup
       - uses: actions/checkout@v2
```
ashleygwilliams:
```suggestion
      - uses: actions/checkout@v3
```

ashleygwilliams: i updated the other workflow but missed this one i guess

docs/book.toml
```diff
@@ -0,0 +1,5 @@
+[book]
+authors = ["Aria Beingessner"]
```
ashleygwilliams:
```suggestion
authors = ["axodotdev"]
```

docs/src/SUMMARY.md
```diff
@@ -0,0 +1,19 @@
+# Summary
+
+- [Introduction](./introduction.md)
+- [Install](./install.md)
```
ashleygwilliams, line 4: should we drop this?

ashleygwilliams (changes requested):
a few small suggestions- also for oranda to pick this up, we'll need to update the oranda.json

docs/src/SUMMARY.md
Gankra: it's discardable 100% but we should at very least link oranda's own generated stuff *from* the book with the heading "install" so people can find it quickly if they end up here

ashleygwilliams: we'll handle book updates post-release, since this is not a release we are formally announcing

ashleygwilliams (approved):

PR #288, opened 2023-05-04
feat(dist): update cargo dist and add npm installer

ashleygwilliams:


.github/workflows/release.yml
```diff
@@ -18,28 +52,33 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Install Rust
-        run: rustup update 1.67.1 --no-self-update && rustup default 1.67.1
+        run: rustup update stable --no-self-update && rustup default stable
```
Gankra, line 55: did you do this on purpose?

Cargo.toml
```diff
@@ -50,9 +50,21 @@ assert_fs = "1.0.7"
 inherits = "release"
 lto = "thin"
 
+# Config for 'cargo dist'
 [workspace.metadata.dist]
-cargo-dist-version = "0.0.5"
+# The preferred cargo-dist version to use in CI (Cargo.toml SemVer syntax)
+cargo-dist-version = "0.0.6-prerelease.8"
+# The preferred Rust toolchain to use in CI (rustup toolchain syntax)
 rust-toolchain-version = "stable"
```
Gankra, line 58: oh huh. pre-existing

Gankra (approved):

.github/workflows/release.yml
ashleygwilliams: i did- mostly because i knew i would forget to and wanted to pin to stable

Cargo.toml
ashleygwilliams: yeah i had edited it

PR #308, opened 2023-05-12
fix a variety of bugs arising from different config states w/repo

ashleygwilliams:


src/site/mod.rs
```diff
@@ -45,48 +45,60 @@ impl Site {
...
+                            content,
+                            &format!("changelog/{}.html", name),
+                            &layout_template,
+                            config,
+                        );
+                        pages.push(page);
+                    }
                 }
+            },
+            None => Err(OrandaError::Other("You have indicated you want to use features that require a repository context. Please make sure you have a repo listed in your project or oranda config."))
```
Gankra: I would prefer that this error message specifically names the key to change, but I guess that's fuzzier with axoproject involved and merging results..? (also the second sentence should use a proper 'help')

src/site/mod.rs
```diff
@@ -45,48 +45,60 @@ impl Site {
...
-                let index = Page::index_with_artifacts(&context, &layout_template, config)?;
-                pages.push(index);
-                if context.latest_dist_release.is_some()
-                    || config.artifacts.package_managers.is_some()
-                {
-                    let body = artifacts::page(&context, config)?;
-                    let artifacts_page =
-                        Page::new_from_contents(body, "artifacts.html", &layout_template, config);
-                    pages.push(artifacts_page);
+        let mut index = Page::index(&layout_template, config)?;
```
Gankra: alternatively you can have index be an Option (= None here) and then at the end you `.unwrap_or_else` with this line, saving the cost of building the page when not needed..?

Gankra (approved):

src/site/mod.rs
ashleygwilliams: updated the message. going through and making sure we're using full miette style errors is gonna be a diff PR

PR #309, opened 2023-05-12
add 'text' as an alias for 'txt' in markdown

Gankra:
This is a super standard convention in the rust ecosystem because defaulting to rust requires a way to say 'no for reals its just text don't do anything fancy'.


ashleygwilliams (approved):
i def approve this, but we should probably also not default to rust highlighting given that oranda supports projects that use all sorts of languages-

ashleygwilliams:
filed https://github.com/axodotdev/oranda/issues/310

PR #316, opened 2023-05-16
feat(mdbook): clean up mdbook integration

### LINKED ISSUE #287: better mdbook integration
We already have a builtin "md_book" key, so we've already blessed it as builtin. We should then ideally only require you to give us the dir where book.toml is and then:

* install mdbook for you (config for version?)
* run `mdbook build` for you
* find the output dir for you (parse book.toml)

Then the entire mdbook section could ideally be removed from the CI, preventing confusing/annoying desyncs between config and CI (alternatively to installing/building mdbook within oranda itself: `oranda init` generates the mdbook stuff in the CI if you enable mdbook integration, so humans can't mess it up).

### LINKED ISSUE #311: mdbook support doesn't work if mdbook's target dir isn't called "book"
This is a confluence of these two pieces of code:

https://github.com/axodotdev/oranda/blob/fe3620595357c7b3322b6b260b0ebd1a03714b6c/src/site/mod.rs#L144

https://github.com/axodotdev/oranda/blob/fe3620595357c7b3322b6b260b0ebd1a03714b6c/src/site/layout/header.rs#L78-L82

The first part copies the dir without renaming it (so whatever name it has ends up in `/public/`

The second part assumes that the resulting dir of the above is `public/book/`.

I'm hitting this because I'm trying to use mdbook-linkcheck in cargo-dist, [and as its docs note](https://github.com/Michael-F-Bryan/mdbook-linkcheck):

>  Note: When multiple [output] items are specified, mdbook tries to ensure that each [output] gets its own sub-directory within the build-dir (book/ by default).
>
> That means if you go from only having the HTML renderer enabled to enabling both HTML and the linkchecker, your HTML will be placed in book/html/ instead of just book/ like before.

So now my book is `../book/book/html/`, which means it becomes `/public/html`, but oranda links `/public/book/`.

### LINKED ISSUE #315: [axomdbook] expand the "mdbook" config to be an Object with many fields
* path to mdbook instance
* whether to enable the axomdbook theme

Gankra:
* config is now called 'mdbook' instead of 'md_book'
* config is now an Object that takes 'path' to the root dir of the book
* we will now build your mdbook for you as part of oranda build
* we will now properly handle when you have multiple renderers (finding the html output)

future work will introduce integrated theme support

fixes #315 
fixes #311
fixes #287

### DIFF
--- a/.github/workflows/web.yml
+++ b/.github/workflows/web.yml
@@ -60,21 +60,7 @@ jobs:
           toolchain: stable
           override: true
 
-      # OPTIONAL: build your mdbook
-      # Note that "./docs/" is the directory in our repo where the book.toml is
-      # and we pass that to mdbook. You also need to tell oranda.json this with
-      #   "md_book": "./docs/book/" 
-      # ("book" being the dir where mdbook writes is output)
-      - name: Install mdbook
-        run: |
-          mkdir mdbook
-          curl -sSL https://github.com/rust-lang/mdBook/releases/download/v0.4.14/mdbook-v0.4.14-x86_64-unknown-linux-gnu.tar.gz | tar -xz --directory=./mdbook
-          echo `pwd`/mdbook >> $GITHUB_PATH
-      - name: Run mdbook build
-        run: |
-          mdbook build docs
-      
-      # Install and run oranda
+      # Install and run oranda (and mdbook)
       # This will write all output to ./public/ (including copying mdbook's output to there)
       - name: Install and run oranda
         # Install from this checkout because WE'RE ORANDA
--- a/docs/book.toml
+++ b/docs/book.toml
@@ -3,3 +3,6 @@ authors = ["axodotdev"]
 language = "en"
 multilingual = false
 src = "src"
+
+[output.markdown]
+[output.html]
--- a/oranda.json
+++ b/oranda.json
@@ -2,7 +2,7 @@
   "favicon": "https://www.axo.dev/favicon.ico",
   "path_prefix": "oranda",
   "mdbook": {
-    "path": "./docs/",
+    "path": "./docs",
     "theme": true
   },
   "changelog": true,
--- a/oranda.json
+++ b/oranda.json
@@ -1,7 +1,10 @@
 {
   "favicon": "https://www.axo.dev/favicon.ico",
   "path_prefix": "oranda",
-  "md_book": "./docs/book/",
+  "mdbook": {
+    "path": "./docs/",
+    "theme": true
+  },
   "changelog": true,
   "social": {
     "image": "https://www.axo.dev/meta_small.jpeg",
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -15,7 +15,7 @@ use std::collections::HashMap;
 
 use theme::Theme;
 
-pub use self::oranda::BookConfig;
+pub use self::oranda::MdBookConfig;
 
 #[derive(Debug)]
 pub struct Config {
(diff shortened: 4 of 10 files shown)


.github/workflows/web.yml
```diff
@@ -60,21 +60,7 @@ jobs:
           toolchain: stable
           override: true
 
-      # OPTIONAL: build your mdbook
```
ashleygwilliams, line 63: you love to see it

oranda.json
```diff
@@ -1,7 +1,10 @@
 {
   "favicon": "https://www.axo.dev/favicon.ico",
   "path_prefix": "oranda",
-  "md_book": "./docs/book/",
+  "mdbook": {
+    "path": "./docs/",
```
ashleygwilliams: is the trailing slash here load bearing? if not i feel like it would be best to remove it

docs/book.toml
```diff
@@ -3,3 +3,6 @@ authors = ["axodotdev"]
 language = "en"
 multilingual = false
 src = "src"
+
+[output.markdown]
+[output.html]
```
Gankra, line 8: oops, i didn't mean to check this in, although it's harmless -- this makes oranda's mdbook more adversarial, placing the output in ./docs/book/html rather than just ./docs/book. Integrating the library makes this transparently handled for us.

oranda.json
```diff
@@ -1,7 +1,10 @@
...
+    "theme": true
```
ashleygwilliams, line 6: i think by default we should theme mdbook using the theme that folks set oranda up with- i think this field should be an override that takes a string that is the theme name (in particular for cases where there is not an mdbook theme for the oranda theme chosen)

src/config/mod.rs
```diff
@@ -37,7 +39,8 @@ pub struct Config {
     pub favicon: Option<String>,
     pub path_prefix: Option<String>,
     pub license: Option<String>,
-    pub md_book: Option<String>,
+    /// The config for using mdbook
+    pub md_book: Option<BookConfig>,
```
ashleygwilliams: i would rename this MdBookConfig as i suspect in the fullness of time we may support other "book" providers

src/site/mod.rs
```diff
@@ -140,11 +142,31 @@ impl Site {
                 full_path.parent().unwrap().as_str(),
             )?;
         }
-        if let Some(book_path) = &config.md_book {
-            Self::copy_static(dist, book_path)?;
+        if let Some(book_cfg) = &config.md_book {
```
ashleygwilliams: if the preferred key name is `mdbook` we should propagate that throughout the program imo

src/site/mod.rs
```diff
@@ -140,11 +142,31 @@ impl Site {
...
+                .expect("mdbook path wasn't utf8");
+
+            // Build the mdbook
+            md.build().map_err(|e| OrandaError::MdBookBuild {
+                path: book_path.clone(),
+                inner: e,
+            })?;
+
+            // Copy the contents to "public/book/"
+            let book_dist = dist.join("book");
```
ashleygwilliams: i suspect that folks will want this to be configurable (there's mdbook config for this ya?)

docs/book.toml
ashleygwilliams: oh interesting, yeah i didn't entirely know what this did- does this mean running mdbook locally would create output in a different location or no?

src/site/mod.rs
Gankra: i agree but i would like to punt on that here, if that's ok? (we already required you to call the dir book, we just broke if you tried to diverge)

ashleygwilliams (approved):

PR #317, opened 2023-05-16
cleanup cargo-dist manifest fetching/selection

### LINKED ISSUE #302: oranda recommends prereleases on the home page
https://opensource.axo.dev/cargo-dist/

is currently recommending v0.1.0-prerelease.3, when ideally it would be recommending v0.7.0, the last stable release.

Gankra:
This makes Release::new async so that we can fetch all Github Releases in parallel.

Because I needed to rework the loop already, I took the time to introduce more robust logic for selecting the 'latest' dist release, in a way that's more aware of non-dist-releases and prereleases.

fixes #302

### DIFF
--- a/src/data/mod.rs
+++ b/src/data/mod.rs
@@ -90,8 +90,10 @@ impl Context {
         // that don't use cargo-dist, we're going to prefer the cargo-dist one, but we should
         // warn the user that things are wonky
         if latest_dist_release.is_some() && latest_dist_release != latest_stable_release {
-            let msg = "You have newer stable Github Releases than your latest cargo-dist Release. Is this intended? (We're going to prefer the cargo-dist one.)";
-            Message::new(MessageType::Warning, msg).print();
+            let dist_rel = &all[latest_dist_prerelease.unwrap()].source.tag_name;
+            let stable_rel = &all[latest_stable_release.unwrap()].source.tag_name;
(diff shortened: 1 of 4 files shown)


src/data/mod.rs
```diff
@@ -33,35 +33,77 @@ impl Context {
...
+        // This prevents us recommending an unstable dist-release over a stable non-dist-release.
+        if latest_stable_release.is_some() {
+            latest_dist_prerelease = None;
+        }
+
+        // If we found a stable cargo-dist release, but there's even newer stable releases
+        // that don't use cargo-dist, we're going to prefer the cargo-dist one, but we should
+        // warn the user that things are wonky
+        if latest_dist_release.is_some() && latest_dist_release != latest_stable_release {
+            let msg = "You have newer stable Github Releases than your latest cargo-dist Release. Is this intended? (We're going to prefer the cargo-dist one.)";
```
ashleygwilliams: i could imagine it'd be useful to say the release names perhaps, e.g. "name@version" - it would make it easier to know if things were correct or not

ashleygwilliams (commented):
lgtm! one comment on improving the error message to be more actionable/auditable

src/data/mod.rs
Gankra: went with tag-name

ashleygwilliams (approved):

PR #326, opened 2023-05-19
reintegrate axoproject

Gankra:


tests/config/mod.rs
```diff
@@ -1,22 +1,24 @@
...
+    let temppath = Utf8Path::from_path(tempdir.path()).expect("non-utf8 temp path");
     let package_json = tempdir.child("package.json");
     package_json
         .write_str(project_config::package_json())
         .expect("failed to write package_json");
 
     assert_eq!(
-        ProjectConfig::detect(&Some(tempdir.path().to_path_buf())),
-        Some(Type::JavaScript(JavaScript {}))
+        ProjectConfig::get_project(temppath).unwrap().0.kind,
```
ashleygwilliams: this is tough to read- could we at least drop the `.0`?

src/errors.rs
```diff
@@ -99,6 +100,14 @@ pub enum OrandaError {
         path: String,
         inner: mdbook::errors::Error,
     },
+    #[error("We found a {kind} project at {manifest_path} but there was an issue")]
```
ashleygwilliams: do we want to add "potential" here like in axoproject?

src/config/project/mod.rs
```diff
@@ -23,35 +19,144 @@ pub struct ProjectConfig {
...
+
+                if let Some(pkg_idx) = package {
+                    // Nice, this package is a perfect candidate
+                    Some((workspace, pkg_idx))
+                } else {
+                    // Found a workspace but none of the packages specifically control this dir.
+                    // This can happen if you run oranda in a dir with a virtual Cargo.toml.
+                    Message::new(
+                        MessageType::Warning,
+                        &format!("Ignoring {:?} project, none of the packages seem to own this directory", workspace.kind),
```
ashleygwilliams: the comment here makes more sense to me than the error message. i think it might be useful to mention that it's likely a cargo workspace and that oranda is per cargo "package" right now and so it should be run in the package dir not the workspace dir

ashleygwilliams (approved):

ashleygwilliams (approved):

PR #328, opened 2023-05-22
fix: add missing logic for non google analytics

### LINKED ISSUE #327: plausible (maybe more analytics) not working
adding the plausible config block does not actually generate any snippets in the resulting build

manyinsects:
closes #327


ashleygwilliams (commented):
lgtm, can we add a test so we dont regress on this again?

manyinsects:
@ashleygwilliams done!

ashleygwilliams (approved):

PR #332, opened 2023-05-22
make the path we pass to mdbook absolute to prevent plugins like link…

Gankra:
…check from freaking out

no idea why this happens. also this is kinda a hacky fix, we should probably do this transform at the time we make the config so that it's relative to oranda.json, and not the cwd.

wanted to put this up before i forget.

### DIFF
--- a/docs/src/configuration/syntax-theme.md
+++ b/docs/src/configuration/syntax-theme.md
@@ -1,6 +1,6 @@
 # Syntax Theme
 
-When it comes to syntax highlighting by default oranda will use the [`material-theme`](./themes/themes.html#materialtheme) and can be easily changed for any of the default themes that can be seen in the [Available themes page](./themes/themes).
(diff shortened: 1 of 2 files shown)


Gankra:
nb this fixes... oh dang i don't think i ever filed the issue

cargo-dist's web.yml doesn't have linkcheck enabled because currently when oranda runs web.yml it just errors out with path errors. this does not occur if you run the actual mdbook cli, so I guess they normalize the path?

https://github.com/axodotdev/cargo-dist/blob/b892c8e3b34eb4cbeb8089c57a350f5e4af46d50/book/book.toml#L12-L13

Gankra:
oh hmm the driveby link fixes i did i think aren't right. i'm not sure if the page this is referring to exists at all anymore

Gankra:
This is now folded into #339

PR #335, opened 2023-05-22
WIP: axomdbook

Gankra:
* adds the contents of https://github.com/axodotdev/axomdbook to oranda-css/mdbook-theme/
* has each build of oranda store those files in the binary
* if `mdbook.theme = true` (or None) then we
   * delete the user's `theme` dir if it exists
   * write_all the files stored in the binary to that dir (creating dirs as needed)
   * toml_edit the their book.toml to set the "axo" theme as the preferred default for html output

The argument for doing these mutations to their project (instead of injecting them with the mdbook library API) is so that `mdbook build` on its own produces the same results. This is currently done on every build, so the user cannot hand-edit the result, as it will just be silently overwritten.
   
image

### DIFF
--- a/docs/theme/css/variables.css
+++ b/docs/theme/css/variables.css
@@ -1,0 +1,359 @@
+
+/* Globals */
+
+:root {
+    --sidebar-width: 300px;
+    --page-padding: 15px;
+    --content-max-width: 750px;
+    --menu-bar-height: 50px;
+    --mono-font: "Source Code Pro", Consolas, "Ubuntu Mono", Menlo, "DejaVu Sans Mono", monospace, monospace;
+    --code-font-size: 0.875em /* please adjust the ace font size accordingly in editor.js */
+}
+
+/* Themes */
+
+.ayu {
+    --bg: hsl(210, 25%, 8%);
+    --fg: #c5c5c5;
+
+    --sidebar-bg: #14191f;
+    --sidebar-fg: #c8c9db;
+    --sidebar-non-existant: #5c6773;
+    --sidebar-active: #ffb454;
+    --sidebar-spacer: #2d334f;
+
+    --scrollbar: var(--sidebar-fg);
+
+    --icons: #737480;
+    --icons-hover: #b7b9cc;
+
+    --links: #0096cf;
+
+    --inline-code-color: #ffb454;
+
+    --theme-popup-bg: #14191f;
+    --theme-popup-border: #5c6773;
+    --theme-hover: #191f26;
+
+    --quote-bg: hsl(226, 15%, 17%);
+    --quote-border: hsl(226, 15%, 22%);
+
+    --table-border-color: hsl(210, 25%, 13%);
+    --table-header-bg: hsl(210, 25%, 28%);
+    --table-alternate-bg: hsl(210, 25%, 11%);
+
+    --searchbar-border-color: #848484;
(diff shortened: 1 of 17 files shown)


Gankra:
This is... arguably shippable as written? just marking it WIP due to my lack of confidence in the approach.

docs/theme/css/variables.css
```diff
@@ -0,0 +1,359 @@
...
+.axo {
+    --color-axo-pink: hsla(326, 100%, 73%, 1);
+    --color-axo-pink-dark: hsla(326, 52%, 58%, 1);
+    --color-axo-orange: hsla(0, 87%, 70%, 1);
+    --color-axo-orange-dark: hsla(356, 75%, 64%, 1);
+    --color-axo-highlighter: hsla(51, 100%, 50%, 1);
+    --color-axo-black: hsla(0, 0%, 5%, 1);
+
+    --title-fg: var(--color-axo-pink);
+    --main-font: Comfortaa;
```
Gankra, line 225: Thankfully mdbook is already heavily variable'd, so 95% of the theme is just settings vars. These are the values we ostensibly want tailwind to inject.

The remaining 5% are:

* a couple more surgical edits to the "real" css to introduce distinctions where we want to make two things that were previously styled the same in mdbook have different styling to more closely match oranda/axo
  * --title-fg and --main-font are new variables that govern these distinctions; all other themes avoid setting these, triggering default values.
* edit to the JS to change the localStorage key for the theme, to avoid conflicts when someone launches a local mdbook that doesn't have the axo theme (causes that other mdbook's JS to crash)
* edit to the JS to tell mdbook that "axo" is a dark theme (affects the JS that does code block syntax highlighting)

Gankra:
See https://github.com/axodotdev/axomdbook/issues for some out-standing issues with the actual theme

See https://github.com/axodotdev/axomdbook/commit/0a052c6ee815461abb195d7a0cd51cf7da108d08 for the precise edits that were made to the default mdbook theme

See https://rust-lang.github.io/mdBook/format/theme/index.html for docs on mdbook themeing

src/site/mdbook.rs
```diff
@@ -0,0 +1,131 @@
...
+        .map_err(|details| {
+            let span = details.span().map(|span| span.into());
+            OrandaError::TomlEdit {
+                source: src.clone(),
+                span,
+                details,
+            }
+        })?;
+    Ok(toml)
+}
```
Gankra, line 129: FIXME: factor into axoasset (cargo-dist also uses a variation of this for editing Cargo.toml in init)

Gankra: https://github.com/axodotdev/axoasset/pull/52

oranda-css/mdbook-theme/css/variables.css
```diff
@@ -0,0 +1,359 @@
...
+
+    --title-fg: var(--color-axo-pink);
+    --main-font: Comfortaa;
+
+    /* modified version of the navy theme above */
+    --bg: var(--color-axo-black);
+    --fg: #bcbdd0;
+
+    --sidebar-bg: var(--color-axo-black);
+    --sidebar-fg: var(--color-axo-orange);
```
Gankra, line 232:
```suggestion
    --sidebar-fg: var(--fg);
```

Suggested on discord, makes the sidebar items grey instead of orange by default

PR #336, opened 2023-05-23
feat: add watch functionality to dev command

manyinsects:
Files that we watch by default:

- `oranda.json` or custom config path
- Readme file (either default or overriden in config)
- mdbook directory (currently not recursively as the HTML output gets put in that directory too)
- Additional pages as specified in the config
- Project manifest files as found by `axoproject`
- User-supplied paths in the `-i, --include-paths` CLI argument

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1172,6 +1172,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 
 [[package]]
+name = "fsevent-sys"
+version = "4.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


manyinsects:
Closes #246

ashleygwilliams (commented):
lgtm! one thing i feel like can be nice is to allow written config for any config you can put on the command line- i don't suspect a ton of people will use it but i could imagine adding a `watch` object to the config, with 2 attrs `include` and `exclude` the both point to a list of paths. this is a nice "get out of jail free" feature for any exceptional cases that come up an we need a temp way to block stuff that allows users to still use the feature- thoughts?

manyinsects:
@ashleygwilliams that does sound like a good idea, but if we do that, we should at the same time think about how we do excludes, since `notify` doesn't really have a mechanism for this. the easiest thing that comes to mind is to `walkdir` down any directories passed in, filter them against an exclude list, and then pass the result to the file watcher... we want this if, for example, people have a `mdbook` structure that includes subfolders, which don't currently get watched by our watcher here

PR #337, opened 2023-05-23
feat: add watch functionality to dev command (for real this time)

manyinsects:
Refer to #336, git sucks so I had to open a new branch.

### DIFF
--- a/src/commands/dev.rs
+++ b/src/commands/dev.rs
@@ -3,6 +3,7 @@ use std::path::PathBuf;
 use axoproject::WorkspaceSearch;
 use camino::Utf8PathBuf;
 use clap::Parser;
+use mdbook::MDBook;
 use notify::{event::ModifyKind, EventKind, Watcher};
 
 use crate::{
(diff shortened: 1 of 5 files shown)


src/commands/dev.rs
```diff
@@ -1,22 +1,150 @@
...
+                config.additional_pages.unwrap().values().cloned().collect();
+            paths_to_watch.append(&mut additional_pages);
+        }
+
+        // Watch for the mdbook directory, if we have it
+        if config.mdbook.is_some() {
+            let path = config.mdbook.unwrap().path;
+            let md = MDBook::load(&path).map_err(|e| OrandaError::MdBookLoad { path, inner: e })?;
+            paths_to_watch.push(md.source_dir().display().to_string());
+        }
```
Gankra:
```suggestion
        if let Some(book_cfg) = &config.mdbook {
            let path = &book_cfg.path;
            let md = MDBook::load(&path).map_err(|e| OrandaError::MdBookLoad { path, inner: e })?;
            
            // watch book.toml and /src/
            paths_to_watch.push(md.root.join("book.toml").display().to_string());
            paths_to_watch.push(md.source_dir().display().to_string());
            
            // If we're not clobbering the theme, also watch the theme dir
            // (note that this may not exist on the fs, mdbook reports the path regardless)
            if !book_cfg.theme.unwrap_or(true) {
                paths_to_watch.push(md.theme_dir().display().to_string());
            }
        }
```

(although ideally watching book.toml would be non-recursive? does that matter?)

manyinsects: afaik `notify` doesn't even consider the recursive setting on files, so it doesn't matter

Gankra (approved):

PR #339, opened 2023-05-23
axomdbook

Gankra:
This adds `./oranda-css/mdbook-theme/` which is vendored into the oranda binary
at build time. If mdbook.theme = true is set (the default, you need to opt out),
we will override your mdbook theme with our own, adding "axo" and "axo-light"
themes which match the default oranda light/dark themes.

The existing builtin mdbook themes should remain and behave the same, and can
still be selected.

We do not modify the source of your mdbook -- we generate a new temp theme dir
in oranda's output dir (./public/) and tell mdbook to use that as the theme.
These themes are currently hardcoded, but use the same css-variable machinery
that mdbook is built around. As such they are primarily defined by a set of
values in variables.css that can easily be updated in the future.

Note that as a result of this, running 'mdbook build' on its own will now
produce different results from having oranda run it. In the future we will
probably expose a "vendor" option that modifies your tree so that vanilla
mdbook theme runs produce the same results (that was actually the original
implementation, but we felt this was less intrusive/surprising as a default).

Some changes made to the default mdbook theme files:

* To avoid issues with running other mdbook instances (or toggling this on/off),
we change the default "mdbook-theme" localStorage key to "axomdbook-theme".
This prevents vanilla instances of mdbook from seeing that you have selected
"axo" as a theme, which would otherwise cause those instances to crash due to
a failed lookup of that name.

* A new highlightjs stylesheet, axoHighlight, has been added to book.js and
index.hbs. Because mdbook doesn't really support adding extra stylesheets
(only overloading the "main" one, which would clash with the default mdbook
themes), we inject this stylesheet by just copying it into the final mdbook
output after mdbook has run. axoHighlight is currently base16/material.css,
which is stored in-tree as
`oranda-css/mdbook-theme/highlight-js-themes/base16-material.css`. It is
selected using your oranda syntax highlighting theme, but that feature was
already limited to one value.

* To avoid issues with adding new highlightjs stylesheets, we now disable all
of them in a loop whenever a theme is selected, so each case only needs to be
responsible for enabling itself, and doesn't need to enumerate all the other
ones that need to be disabled.

* A few edits were made to the css to add more distinctions-via-variables,
like allowing "main titles" to be styled differently from other headings,
and allowing themes to select the font.

* axo/axo-light are hardcoded in the css to disable padding/bg-color-changes
to inline code snippets, as this is the style we use in oranda. I couldn't
find a way to make this controlled by variables, as this is specifically
shadowing values set by the currently enabled highlightjs stylesheet, and
there's no way to "disable" a css rule with just variables (that I know).

* fonts.css was changed to import axo's preferred fonts instead of the
builtin ones. I'm a bit fuzzy on whether the vendored fonts are unvendored
by this, because this behaviour was recently being changed in mdbook.


Gankra:
alright, did a bunch of work on this and it's looking a lot better.

Some issues:

* standard mdbook themes don't have their fonts vendored anymore so they get console errors (left a FIXME in fonts.css)
* standard mdbook themes lose the bg/padding around inline code (left a FIXME in variables.css)
* our quote blocks don't really match anything (oranda and blog.axo.dev completely disagree on the style)

src/site/mdbook.rs
```diff
@@ -0,0 +1,145 @@
...
+    book_cfg: &MdBookConfig,
+    syntax_theme: &SyntaxTheme,
+) -> Result<()> {
+    Message::new(MessageType::Info, "Building mdbook...").print();
+    tracing::info!("Building mdbook...");
+
+    // Read mdbook's config to find the right dirs
+    let book_dir = Utf8PathBuf::from(&book_cfg.path);
+
+    if book_cfg.theme.unwrap_or(true) {
```
manyinsects: Can this be `Some(false)` or does this condition always pass? As in, is there a chance someone's gonna put in `theme: false` in their config?

src/site/mdbook.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    // Copy the contents to "public/book/"
+    // FIXME: make this something they can set in the MdBookConfig
+    let book_dist = dist.join("book");
+    Site::copy_static(&book_dist, build_dir.as_str())?;
+
+    Ok(())
+}
+
+fn load_mdbook(book_dir: &Utf8Path) -> Result<MDBook> {
```
manyinsects: do you think it'd be a good idea to `pub` this function so we can use it in other places (like the dev command, where we also need to load the mdbook configuration)?

Gankra: 100%

src/site/mdbook.rs
Gankra: yes, theme = false is basically the only value we expect people to manually set (or in the future, a theme name, maybe).

Gankra:
wait what

manyinsects (approved):
looking good, i think this can be merged after reverting the web.yml change and squashing

Issue #282, opened 2023-05-04
Support pretty links


manyinsects:
Right now, we output all HTML files as just that - plain HTML files, which ends up meaning we have to link to them like this: `https://myorandasite.cool.gay/changelog.html`. If we moved HTML files into their own folder (e.g. `changelog/index.html` instead of `changelog.html`), we would lose the `.html` extension, which, AFAIK, is important for SEO (and probably a bunch of other things as well). This is sort of a default feature for a 21st-century-SSG, so I think it should be on by default, maybe with an option to disable it (although I don't see much of a point in that).

PR #351, opened 2023-05-31
Write "pretty links" to destination directory

manyinsects:
Instead of writing `changelog.html` (for example), we want to write `changelog.html` so that the end user can generate "pretty links", i.e. linking to `https://mysite.com/changelog`. This PR Does That.

I ended up being extra pedantic about creating links wit h a trailing slash, because some links didn't, and therefore broke when the user supplied a custom path prefix (when you link to `/mypathprefix/changelog`, the browser can interpret it as linking to `/changelog` for some reason?).

I also updated build tests to use raw string literals for any HTML strings we were testing for, since that saves us a bunch of having to escape quotation marks.

### DIFF
--- a/src/site/mod.rs
+++ b/src/site/mod.rs
@@ -133,10 +133,17 @@ impl Site {
     pub fn write(self, config: &Config) -> Result<()> {
         let dist = Utf8PathBuf::from(&config.dist_dir);
         for page in self.pages {
-            // FIXME: We have to do some gymnastics here due to `LocalAsset::write_new_all` taking filename and dest
-            // path separately. Hopefully in a future version of axoasset, this only takes one parameter instead of
-            // two, and we can just add the page filename to the dest path and pass it in.
-            let full_path = dist.join(&page.filename);
+            let filename_path = Utf8PathBuf::from(&page.filename);
+            // Prepare to write a "pretty link" for pages that aren't index.html already. This essentially means that we rewrite
+            // the page from "page.html" to "page/index.html", so that it can be loaded as "mysite.com/page" in the browser.
+            let full_path: Utf8PathBuf = if !filename_path.ends_with("index.html") {
+                // FIXME: Can we do anything BUT unwrap here? What's the smart way to deal with a missing filename path portion?
+                let file_stem = filename_path.file_stem().unwrap();
+                let parent = filename_path.parent().unwrap_or("".into());
+                dist.join(parent).join(file_stem).join("index.html")
+            } else {
+                dist.join(filename_path)
+            };
             LocalAsset::write_new_all(&page.contents, full_path)?;
         }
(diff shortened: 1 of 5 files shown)


manyinsects:
Closes #282!

src/site/mod.rs
```diff
@@ -133,10 +133,17 @@ impl Site {
...
-            // FIXME: We have to do some gymnastics here due to `LocalAsset::write_new_all` taking filename and dest
-            // path separately. Hopefully in a future version of axoasset, this only takes one parameter instead of
-            // two, and we can just add the page filename to the dest path and pass it in.
-            let full_path = dist.join(&page.filename);
+            let filename_path = Utf8PathBuf::from(&page.filename);
+            // Prepare to write a "pretty link" for pages that aren't index.html already. This essentially means that we rewrite
+            // the page from "page.html" to "page/index.html", so that it can be loaded as "mysite.com/page" in the browser.
+            let full_path: Utf8PathBuf = if !filename_path.ends_with("index.html") {
+                // FIXME: Can we do anything BUT unwrap here? What's the smart way to deal with a missing filename path portion?
+                let file_stem = filename_path.file_stem().unwrap();
```
Gankra, line 141: This can only be None if filename_path is the empty string, I believe. Unwrapping is extremely reasonable.

Gankra (approved):

tests/build/mod.rs
```diff
@@ -232,7 +232,8 @@ fn adds_changelog_nav() {
     let config = oranda_config::changelog(temp_dir);
     let layout = Layout::new(&config).unwrap();
     let page = page::index(&config, &layout);
-    assert!(page.contents.contains("changelog.html"));
+    dbg!(&page.contents);
```
ashleygwilliams: do we want to leave this dbg here?

ashleygwilliams (commented):
i think in general this looks good. i am interested in the trailing slash and exactly how that works. probably not something to block this- but i want to better understand the plan from an end user perspective

tests/build/mod.rs
manyinsects: oops. no

manyinsects:
@ashleygwilliams not sure on the actual internals of it, but I noticed that a lot of the time in my local dev setup, I was getting a link to, for example, `/oranda/artifacts` to redirect to `/artifacts`. the HTML link was generated correctly, but for some reason my browser scrubbed away the path prefix. adding a trailing slash fixed it, and I think it's only a problem for links from within the same domain/site

ashleygwilliams (approved):
i think we should land this :) i do think we will want to potentially figure out how to make trailing slashes non-load bearing but we can do this separately (if it is just an artifact of oranda dev that's one thing, but i think in prod that will be frustrating for folks)

PR #352, opened 2023-05-31
fix: check for file/folder existence before watching

### LINKED ISSUE #341: `oranda dev` expects `oranda.json` to exist
This is due to how `notify` expects paths given to it to exist (which makes total sense). We should probably implement a `Path::exists` check for every path we pass to `notify`.

manyinsects:
Closes #341.

### DIFF
--- a/src/commands/dev.rs
+++ b/src/commands/dev.rs
@@ -125,6 +125,8 @@ impl Dev {
 
         for path in paths_to_watch {
             let path = PathBuf::from(path);
+            // If no path exists, oranda won't work anyways, so there's not much need to let the user know
+            // (they'll know either way ;) )
             if path.exists() {
                 watcher.watch(path.as_path(), notify::RecursiveMode::Recursive)?;
             }
--- a/src/commands/dev.rs
+++ b/src/commands/dev.rs
@@ -125,7 +125,9 @@ impl Dev {
 
         for path in paths_to_watch {
             let path = PathBuf::from(path);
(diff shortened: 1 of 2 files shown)


Gankra (approved):
It would be cool if we could detect the dirs getting created but that sounds like a lot more messy work for minimal benefit.

src/commands/dev.rs
```diff
@@ -125,7 +125,9 @@ impl Dev {
 
         for path in paths_to_watch {
             let path = PathBuf::from(path);
-            watcher.watch(path.as_path(), notify::RecursiveMode::Recursive)?;
+            if path.exists() {
+                watcher.watch(path.as_path(), notify::RecursiveMode::Recursive)?;
+            }
```
ashleygwilliams, line 132: do we want to give a message if it doesn't that helps explain what might be happening?

manyinsects: not sure how helpful that would be, imagine you have a minimal oranda setup (no config file, no book, etc), and you immediately get like 3 warning log messages. since the user doesn't add these paths i think it's fine to silently ignore

ashleygwilliams: ah, so i was not fully understanding what was happening here. now that i do - i can make my comment more specific, which is... do we warn/error if there are no paths being watched? (is it possible that no path can be watched?)

manyinsects: yes, if the user doesn't even have a readme file, although i think oranda will error anyways in that case

PR #355, opened 2023-06-01
fix: bad config docs link in the README

### LINKED ISSUE #354: README configuration section bad link
Hello,

The link under the configuration header is pointing at https://opensource.axo.dev/oranda/configuration which does not seem to be a valid link.

I believe https://opensource.axo.dev/oranda/book/configuration.html is the correct page.

gd0st:
Fixes axodotdev/oranda#354

Fixed the README to point at https://opensource.axo.dev/oranda/book/configuration.html instead of its currently broken link.


manyinsects (approved):
Thank you for catching that!

PR #356, opened 2023-06-02
feat(auto-detect): add basic auto-detection for cargo-dist support

Gankra:


src/config/project/mod.rs
```diff
@@ -15,6 +15,7 @@ pub struct ProjectConfig {
     pub repository: Option<String>,
     pub version: Option<String>,
     pub license: Option<String>,
+    pub readme_path: Option<Utf8PathBuf>,
```
ashleygwilliams, line 18: can you say more about adding this here? does this connect to axoproject? (it's not clear to me we ever receive this value when deserializing a project manifest so i want to make sure it's connected)

ashleygwilliams (commented):
i am not entirely sure if the refactor maintains all the previous logic, as somethings have defaults and others don't. is this still using defaults where available?

src/config/project/mod.rs
Gankra: It's a value that Cargo.toml lets you specify and axoproject already reads. I did a quick pass of "are there values we're not exposing that we already have equivalents of in oranda" and this is the only one that turned up.

ashleygwilliams: does axoproject do it for js files as well? package.json has a readme key - i am cool with adding this, just want to make sure that it's fully baked :)

manyinsects (approved):

PR #357, opened 2023-06-02
refactor config/mod.rs to apply layers more cleanly

Gankra:


ashleygwilliams (commented):
is this the same commit that is in the previous pr for auto-detection? they seem nearly identical...

Gankra:
Yeah it's rebased on top of this branch

Gankra:
landed in #356

PR #358, opened 2023-06-05
feat: add funding page

### LINKED ISSUE #199: support funding.yml and oranda.json funding config

manyinsects:
Closes #199, and supersedes #221. Adds a `funding` config key that, when set to true, will try and fetch from the `repository` to see if there's a `FUNDING.yml` file. Screenshot with almost every available key:

CleanShot 2023-06-08 at 23 40 44@2x

### Testing locally

- Make sure that you're using oranda-css local
- In your oranda checkout, create a `.github/FUNDING.yml` file and fill it with some keys (see an example [here](https://github.com/shadows-withal/example-funding-repo/blob/main/.github/FUNDING.yml))
- Optionally create a `funding.md` and/or set `funding.preferred_funding` in your config

### Outstanding things

- [x] We might need SVG icons for IssueHunt and Otechie, but I'm no good with mangling raw SVG, maybe @SaraVieira could do this?
- [x] How do we want to behave if we have funding enabled but there's no Funding file? The version right now errors out with a diagnostic tip.
- [x] Do we want the page to be more descriptive than just showing buttons with hover tooltips?
- [x] Do we want to link to this page from anywhere at all?


SaraVieira:
> We might need SVG icons for IssueHunt and Otechie, but I'm no good with mangling raw SVG, maybe @SaraVieira could do this?
Otechie has died and one for IssueHunt has been sent

> How do we want to behave if we have funding enabled but there's no Funding file? The version right now errors out with a diagnostic tip.

In my opinion we should skip it and show a warning, an error seems a bit harsh

> Do we want the page to be more descriptive than just showing buttons with hover tooltips?

After playing with it for a while I think something like this for all buttons would be nice
CleanShot 2023-06-05 at 12 30 57@2x

ashleygwilliams (commented):
thinking about the type of content to add to this to potentially help make it easier to layout...

- "preferred or featured" funding option
- funding.md - content to explain the funding.yml

SaraVieira:
New layout pushed
CleanShot 2023-06-05 at 16 45 09@2x

ashleygwilliams (commented):
> Do we want the page to be more descriptive than just showing buttons with hover tooltips?

I think we should support a "magic" funding.md (similar treatment to readme.md, expected to be top-level, can config path to) that is rendered on the top of this page. i also think we should give a "preferred" config option and highlight that funding option in a larger display at the top of the page. this gives us a layout of:

[preferrred funding]
[contents of funding.md]
[list of all funding options]

> Do we want to link to this page from anywhere at all?

We should add it as a top level element in the nav.

ashleygwilliams (commented):
i also think it might be nice to do autodetection for this (cc @gankra) so that the config is to set it to `false` as opposed to `true` in the presence of a `funding.yml` - this addresses the "what to do if it is set to true and it is not there" question a bit as people are not incentivized to config true, but rather false.

src/config/oranda_config.rs
```diff
@@ -50,6 +51,12 @@ impl Default for StyleConfig {
     }
 }
 
+/// Config for displaying funding information on your page
+#[derive(Debug, Default, Deserialize)]
+pub struct FundingConfig {
+    pub preferred_funding: Option<FundingType>,
```
ashleygwilliams, line 83: do we want to offer configuring the path to the funding.md?

src/data/funding.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/// Loads a `funding.md` file from the root directory, to serve as documentation
+/// for the generated funding page. Returns the same as the above function.
+pub fn load_funding_docs() -> Result<Option<String>> {
+    // FIXME: Do we want this file to be FUNDING.md?
+    load_generic_file("funding.md")
+}
+
+fn load_generic_file(path: &str) -> Result<Option<String>> {
```
ashleygwilliams: is this functionality we can put in axoasset to make it easier to use? i'd hope we wouldn't need to wrap like this

src/errors.rs
```diff
@@ -141,6 +141,17 @@ pub enum OrandaError {
...
 
+    #[error("Failed to fetch your funding info from GitHub.")]
+    #[diagnostic(help = "Make sure that your funding file is located at `.github/FUNDING.yml`.")]
+    GithubFundingFetchError {
+        #[source]
+        details: reqwest::Error,
+    },
+
+    #[error("Error while parsing FUNDING.yml")]
+    #[diagnostic(help = "Make sure your FUNDING.yml conforms to GitHub's format!")]
```
ashleygwilliams: is there a doc on this we can link to?

ashleygwilliams (commented):
looks good, some smaller things

src/errors.rs
manyinsects: yes, good idea!

PR #359, opened 2023-06-05
feat(auto-mdbook): add autodetection for mdbook

Gankra:
All of the following are now valid values for the mdbook config:

* (no key) - enables mdbook, autodetects path, custom theme
* mdbook: true - redundant synonym for the above
* mdbook: false - disables mdbook
* mdbook: { path: ./doc/ } - enables mdbook with an explicit path, custom theme
* mdbook: { theme: false } - enables mdbook, autodetects path, disables custom theme

### DIFF
--- a/oranda.json
+++ b/oranda.json
@@ -1,10 +1,6 @@
 {
   "favicon": "https://www.axo.dev/favicon.ico",
   "path_prefix": "oranda",
-  "mdbook": {
-    "path": "./docs",
-    "theme": true
-  },
   "changelog": true,
   "social": {
     "image": "https://www.axo.dev/meta_small.jpeg",
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -13,6 +13,8 @@ use oranda_config::{OrandaConfig, Social};
 use project::ProjectConfig;
 use std::collections::HashMap;
 
+use self::oranda_config::BoolOr;
+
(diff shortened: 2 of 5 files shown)


Gankra:
(also based on top of #357)

ashleygwilliams (commented):
do you see a path for auto-custom-themeing? i think a common situation will be that folks will want a custom theme that matches their oranda theme. i think making folks have to specify a path and a theme to do that seems like a lot if it is otherwise autodetected- i know @SaraVieira is working on the other themes stuff right now- can we have it theme your mdbook the same as your page (aka default to axo) so that folks only have to write config to turn that off?

Gankra:
The mdbook code already looks at what oranda uses for the theme and uses the same thing (if it has a mapping available).

https://github.com/axodotdev/oranda/blob/92ad30582a94315d774f4951e533e12b6efa0bc8/src/site/mdbook.rs#L46-L54

https://github.com/axodotdev/oranda/blob/92ad30582a94315d774f4951e533e12b6efa0bc8/src/site/mdbook.rs#L95-L102

https://github.com/axodotdev/oranda/blob/92ad30582a94315d774f4951e533e12b6efa0bc8/src/site/mdbook.rs#L124-L127

ashleygwilliams (approved):
ah ok cool, i misremembered where we landed on that. this looks good to me

PR #363, opened 2023-06-06
chore: make configs more optional and mergeable

### LINKED ISSUE #340: Error when changing theme
Trying to change the theme in the oranda.json gives me the following error:

image

Looking at the struct it seems that `theme`, `syntax_theme` and `additional_css` are required when inside the config nothing should be required since using any of these would be a valid config

```rs
pub struct StyleConfig {
    pub theme: Theme,
    pub syntax_theme: SyntaxTheme,
    pub additional_css: Vec<String>,
    pub oranda_css_version: Option<String>,
}
```

Gankra:
fixes #340

### DIFF
--- a/src/config/artifacts.rs
+++ b/src/config/artifacts.rs
@@ -37,6 +37,17 @@ pub struct Artifacts {
 }
 
 impl Artifacts {
+    /// Merge this value with another layer of itself, preferring the new layer
+    pub fn apply_layer(&mut self, layer: Self) {
+        if let Some(val) = layer.cargo_dist {
+            self.cargo_dist = Some(val);
+        }
+        if let Some(val) = layer.package_managers {
+            // FIXME: should this get merged with e.g. `extend?`
+            self.package_managers = Some(val);
+        }
+    }
+
     pub fn has_some(&self) -> bool {
         self.cargo_dist() || self.package_managers.is_some()
     }
--- a/src/config/artifacts.rs
+++ b/src/config/artifacts.rs
@@ -1,6 +1,8 @@
 use linked_hash_map::LinkedHashMap;
 use serde::Deserialize;
 
+use super::{ApplyLayer, ApplyOptExt};
+
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "lowercase")]
 enum ArtifactSystem {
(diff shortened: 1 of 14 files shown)


manyinsects (approved):
we're defining `apply_layer` so many times that it really feels like it should be a trait, but that's not a blocker imo

Gankra:
I don't disagree but also I don't know what purpose the trait would actually serve -- I suppose streamlining the Option handling?

src/config/artifacts.rs
```diff
@@ -36,6 +38,14 @@ pub struct Artifacts {
     pub package_managers: Option<LinkedHashMap<String, String>>,
 }
 
+impl ApplyLayer for Artifacts {
+    fn apply_layer(&mut self, layer: Self) {
+        self.cargo_dist.apply_opt(layer.cargo_dist);
+        // FIXME: should this get merged with e.g. `extend?`
```
ashleygwilliams, line 44: where is the other location that a user can specify package_managers?

src/config/mod.rs
```diff
@@ -77,107 +77,39 @@ impl Config {
...
+            self.description.apply_val(custom.description);
+            self.dist_dir.apply_val(custom.dist_dir);
+            self.static_dir.apply_val(custom.static_dir);
+            self.homepage.apply_opt(custom.homepage);
+            self.name.apply_val(custom.name);
+            self.readme_path.apply_val(custom.readme_path);
+            self.repository.apply_opt(custom.repository);
+            self.analytics.apply_layer(custom.analytics);
+            // FIXME: should this get merged with e.g. `extend?`
+            self.additional_pages.apply_opt(custom.additional_pages);
```
ashleygwilliams, line 103: where is the other place that additional pages could be specified?

src/config/oranda_config.rs
```diff
@@ -30,23 +40,39 @@ pub struct MdBookConfig {
...
+    fn apply_layer(&mut self, layer: Self) {
+        self.theme.apply_opt(layer.theme);
+        self.syntax_theme.apply_opt(layer.syntax_theme);
+        self.oranda_css_version.apply_opt(layer.oranda_css_version);
+        self.additional_css.extend(layer.additional_css);
+    }
+}
+impl StyleConfig {
+    /// Get the theme
+    pub fn theme(&self) -> Theme {
```
ashleygwilliams, line 70: why are we adding functions here instead of using the default values as values?

ashleygwilliams (commented):
a few questions - 

in some places, you comment about extending vs clobbering- but it's not clear to me how a user is specifying those keys in different places... (they are only available in oranda.json as far as i know). my personal preference is clobber because it keeps the source of truth singular fo r each key and allows you to "turn off" things that may be specified in other files that you don't have full control over and/or want oranda to ignore. i am a little concerned though because in general those keys should already only be specified once.

also not clear why theme and syntax_theme are functions now- for a few reasons i think we should treat them as values. curious to understand more

ashleygwilliams:
i am gonna merge this to avoid conflicts, but @Gankra it'd be good to handle those FIXMEs on monday

src/config/artifacts.rs
Gankra: Nowhere, but each layer """has to""" assume that other layers can set these values, as a matter of like, robust abstraction / consistency?

src/config/mod.rs
Gankra: ditto

src/config/artifacts.rs
Gankra: These FIXMEs are more like "hey this code might make your sad if we add new layers that touch these values".

src/config/oranda_config.rs
Gankra: This is the crux of a discussion I brought up with @shadows-withal last week.  Some values in Theme should be Optional in OrandaConfig, but aren't Optional in the final Config. This raises a question: "Do we want the final Config to use the same struct for things like StyleConfig that OrandaConfig uses?" (Before my PR the answer was "yes" and after my PR it remains "yes")

* If no, then we can just have different Optionality and everything is clean, but now you are defining ~everything multiple times
* If yes, then users of the final Config need to deal with the fact that the value is Optional, even though it isn't. I chose to deal with this fact by removing the default value from the Default impl, and moving it to a method that wraps them up in a method that returns the default value.

PR #379, opened 2023-06-14
feat: restyle default oranda themes

manyinsects:
Dark (default) theme screenshot:

image

Light theme screenshot:

image


ashleygwilliams (commented):
q: does this preserve the axo themes as a separate theme or overwrite them? i think we'd want to keep the "original" as a theme for at least ourselves

ashleygwilliams:
Screenshot 2023-06-14 at 7 47 28 AM

my suspicion is that this won't pass a11y checks - we may want to change one or both of the colors here

manyinsects:
> q: does this preserve the axo themes as a separate theme or overwrite them? i think we'd want to keep the "original" as a theme for at least ourselves

Yeah, we should add an "axo" theme, which I'll do after I've adjusted the mdbook themes.

src/site/mdbook.rs
```diff
@@ -128,7 +142,9 @@ impl AxomdbookTheme {
     pub fn name(&self) -> &'static str {
         use AxomdbookTheme::*;
         match self {
-            Axo => "Axo Dark",
+            Default => "Default Dark",
+            DefaultLight => "Default Light",
```
Gankra: These should probably be "Oranda Dark" and "Oranda Light"

oranda-css/mdbook-theme/oranda-themes/default.css
```diff
@@ -0,0 +1,133 @@
+.oranda-dark {
+    /* 
+      This part is just defining constants that are consistent
+      between both axo themes, which oranda-css should probably
```
Gankra:
```suggestion
      between both oranda themes, which oranda-css should probably
```

oranda-css/mdbook-theme/oranda-themes/default.css
```diff
@@ -0,0 +1,133 @@
...
+    --searchresults-header-fg: var(--title-fg);
+    --searchresults-border-color: var(--border-color);
+    --searchresults-li-bg: var(--well-bg);
+    --search-mark-bg: var(--well-bg);
+}
+
+.oranda-light {
+    /* 
+      This part is just defining constants that are consistent
+      between both axo themes, which fringe/oranda-css should probably
```
Gankra:
```suggestion
      between both oranda themes, which fringe/oranda-css should probably
```

oranda-css/mdbook-theme/fonts/fonts.css
```diff
@@ -1,5 +1,6 @@
 /* fonts used by axo themes, not sure if this is the best way to fetch them */
 
+@import url("https://fonts.googleapis.com/css2?family=Fira+Sans:wght@400;700&display=swap");
```
Gankra, line 3: Oh right, I forgot to look into making this part more properly customizeable per-theme. One option that I think *might* work is to include these lines at the top of each theme variables file (the files where you declare `.oranda-dark`). Not something you need to fix, kinda just mumbling at myself here.

Gankra (commented):
Only reviewed the mdbook parts, a few nits.

oranda-css/mdbook-theme/fonts/fonts.css
manyinsects: that does sound like a better solution for sure

ashleygwilliams: @Gankra do you want to file an issue for this - i want to try to get this PR merged but i don't want to lose this idea as it sounds like you both prefer it

src/site/mdbook.rs
ashleygwilliams: i feel complicated about this- i sorta like default because it helps me know it's the default- can you say more about why you prefer oranda? (i can imagine way, but am curious to hear you articulate it)

Gankra: because this is showing up in the dropdown menu where you select themes, and it's not The Default For Mdbook Itself, so it's a little weird.

Gankra: (nb mdbook already has themes called "dark" and "light")

ashleygwilliams: ahhhh yeah, i was forgetting that we still show all the themes (ya?) if so then yeah we should update this

oranda-css/mdbook-theme/fonts/fonts.css
ashleygwilliams: https://github.com/axodotdev/oranda/issues/383

ashleygwilliams (commented):
~~so currently, if i use the default light/dark it will also default my mdbook to that theme. however, if i use axodark, it doesn't default the styles to axodark on mdbook, tho the option is listed as available. need to figure out how to make that work. (behavior is fine for all themes except the axo ones...)~~

EDIT: i ran into a pre-existing footgun that @Gankra was aware of- this PR didn't intro it

ashleygwilliams (approved):

PR #382, opened 2023-06-14
feat(custom-installers): create a new middle layer for artifacts

Gankra:
* [x] write middle layer
  * [x] cargo-dist
  * [x] inference
  * [x] package-manager  
  * [x] harmonize data sources to feel uniform 
* [x] port downloads table
  * [x] fix `<th>` 1rem padding thing in oranda-css
* [x] port installer widget
   * [x] new JS/html that interprets ReleaseArtifacts (with @shadows-withal)
   * [x] style said html nicely
   * [x] better message for "we actually did detect your OS but no matching platform"
   * [x] add back link to full downloads page
   * [x] add simpler modes for simple cases
 * [x] add back verbose install page listings (use new config)
 * [x] update snapshot tests


src/data/artifacts.rs
```diff
@@ -0,0 +1,235 @@
...
+                    .then_with(|| installer_a.label.cmp(&installer_b.label))
+            });
+
+            let installers = installers.into_iter().map(|(i, _pref)| i).collect();
+            self.targets.insert(target.to_owned(), installers);
+        }
+    }
+}
+
+/// Take an installer preference an uniformly apply it to every given TargetTriple
```
manyinsects:
```suggestion
/// Take an installer preference and uniformly apply it to every given TargetTriple
```

manyinsects (commented):
looks good as a backbone for this stuff!

src/data/artifact_inference.rs
```diff
@@ -0,0 +1,352 @@
+use camino::Utf8PathBuf;
+
+use super::artifacts::*;
+
+// Architectures
```
ashleygwilliams: are these commented out on purpose?

src/data/artifact_inference.rs
```diff
@@ -0,0 +1,352 @@
+use camino::Utf8PathBuf;
+
+use super::artifacts::*;
```
ashleygwilliams: i think artifacts should be a dir and this file should be inside it

src/data/artifact_inference.rs
```diff
@@ -0,0 +1,352 @@
...
+// const ARCH_ARM64: &str = "aarch64";
+
+// OSes
+// const OS_WINDOWS: &str = "pc-windows-msvc";
+// const OS_LINUX_GNU: &str = "unknown-linux-gnu";
+// const OS_LINUX_MUSL: &str = "unknown-linux-musl";
+// const OS_MAC: &str = "apple-darwin";
+
+// Various target triples
+const TARGET_X86_WINDOWS: &Targ = "i686-pc-windows-msvc";
```
ashleygwilliams: i think it would be useful to put the constants in a different file potentially

src/data/artifact_inference.rs
```diff
@@ -0,0 +1,352 @@
...
+                }
+            }
+
+            let label;
+            let description = String::new();
+            let method;
+            let preference;
+
+            // Try to detect what kind of file this is
+            if file.name.contains("install")
```
ashleygwilliams: could we break these conditions into individual functions?

src/data/cargo_dist.rs
```diff
@@ -1,65 +1,112 @@
...
-        _ => String::from("Unknown"),
-    }
-}
+                // If this artifact has a checksum, register it
+                let checksum_file = artifact.checksum.as_ref().and_then(|n| self.file_idx(n));
+                if let Some(file) = file {
+                    self.file_mut(file).checksum_file = checksum_file;
+                }
+
+                match artifact.kind {
```
ashleygwilliams, line 39: could we break these out into different functions?

src/data/github/mod.rs
```diff
@@ -74,3 +76,19 @@ impl GithubRelease {
         None
     }
 }
+
+impl ReleaseArtifacts {
```
ashleygwilliams, line 80: it is somewhat surprising that this impl is here- i also get why it's here but i wonder if there's not a better way to structure things to make it easier to find

src/data/mod.rs
```diff
@@ -14,43 +16,78 @@ pub struct Context {
     pub repo: GithubRepo,
     pub releases: Vec<Release>,
     pub has_prereleases: bool,
-    pub latest_dist_release: Option<DistRelease>,
+    pub latest_release: Option<usize>,
```
ashleygwilliams, line 26: i am surprised this is a type usize...

src/data/mod.rs
```diff
@@ -14,43 +16,78 @@ pub struct Context {
...
+    /// * whether there are any prereleases (enables more complex UI)
+    /// * whether there are any artifacts (enables more complex UI)
+    /// * the release that should be considered "latest" (shown on the front page)
     #[allow(clippy::unnecessary_unwrap)]
     pub fn fetch_all_releases(
         repo: &GithubRepo,
-        cargo_dist: bool,
-    ) -> Result<(Vec<Release>, bool, Option<DistRelease>)> {
+        artifacts_config: &ArtifactsConfig,
+    ) -> Result<(Vec<Release>, bool, bool, Option<usize>)> {
```
ashleygwilliams: this return type is starting to look like it wants to be a struct... 2 bools seems like it'll get confusing

src/site/artifacts/installers.rs
```diff
@@ -1,218 +1,285 @@
...
+            <h4>{text!("Install {}", release.source.tag_name)}</h4>
+            <div><small class="published-date">{text!("Published on {}", formatted_date)}</small></div>
+
+            <ul class="arches">
+                {arches}
+            </ul>
+        </div>
+    );
+
+    // If there's only one platform we don't need scripts
```
ashleygwilliams, line 45: could we break these out into smaller functions?

ashleygwilliams (approved):
mostly requests for smaller functions and to re-org some of the file structure in artifacts. we can merge this and i can do that re-org if you'd like

src/data/artifact_inference.rs
Gankra: Yeah they might be needed in the future, and wanted to leave them for reference for the js.

src/data/mod.rs
Gankra: it's an index into an array

src/data/mod.rs
Gankra: yeah this function is a disaster

src/data/mod.rs
ashleygwilliams: oooooooooooohhhhh- thoughts on call it latest_release_index?

ashleygwilliams (commented):
🚀

Issue #392, opened 2023-06-16
lists in paragraphs should have bullet points/numbers


ashleygwilliams:
<img width="262" alt="Screenshot 2023-06-16 at 6 01 05 PM" src="https://github.com/axodotdev/oranda/assets/1163554/406b750b-db5e-4030-935d-aa4811440115">

Issue #387, opened 2023-06-16
images are not centered


marcoieni:
I'm trying oranda in [this](https://github.com/MarcoIeni/release-plz/pull/826/files) branch, and I noticed images are not centered 
![image](https://github.com/axodotdev/oranda/assets/11428655/64a2a29f-8ede-41f5-bde0-3f313595d78e)

manyinsects:
thanks for the report! that's definitely something we'll fix asap

PR #399, opened 2023-06-19
Various CSS fixes

### LINKED ISSUE #253: changelog page is too narrow, makes reading unpleasant

manyinsects:
Closes #253, #205, #392, and #387.

### DIFF
--- a/oranda-css/css/base.css
+++ b/oranda-css/css/base.css
@@ -128,6 +128,11 @@ li {
   @apply list-none;
 }
 
+.rendered-markdown ul,
+.rendered-markdown li {
+  @apply list-disc;
+}
+
 li {
   @apply sm:ml-8 ml-4 mb-4;
 }
--- a/oranda-css/css/base.css
+++ b/oranda-css/css/base.css
@@ -178,7 +178,7 @@ blockquote {
 }
 
 main {
-  @apply mx-auto my-24 lg:max-w-3xl max-w-[80%];
+  @apply mx-auto my-24 lg:max-w-4xl max-w-[80%];
 }
 
 .github-icon {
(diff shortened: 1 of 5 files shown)


src/site/page/mod.rs
```diff
@@ -71,7 +73,14 @@ impl Page {
     fn load_and_render_contents(source: &str, syntax_theme: &SyntaxTheme) -> Result<String> {
         let source = SourceFile::load_local(source)?;
         let contents = source.contents();
-        markdown::to_html(contents, syntax_theme)
+        markdown::to_html(contents, syntax_theme).map(|html| {
+            let html: Box<div<String>> = html!(
+                <div class="rendered-markdown">
```
ashleygwilliams, line 78: this is a great call

oranda-css/css/base.css
```diff
@@ -178,7 +179,7 @@ blockquote {
 }
 
 main {
-  @apply mx-auto my-24 lg:max-w-3xl max-w-[80%];
+  @apply mx-auto my-24 lg:max-w-4xl max-w-[80%];
```
ashleygwilliams, line 182: i dont think i'm seeing this change on my screen width and i think it would be useful there - can you send a screen cap of what it looks like so i can compare?

manyinsects: new:

CleanShot 2023-06-19 at 15 29 22@2x

old:

CleanShot 2023-06-19 at 15 29 55@2x

ashleygwilliams: what width does it need to trigger this? (def looks better!)

manyinsects: i think the breakpoint is `lg`, which comes out to about 1024px

ashleygwilliams: i'm dumb and wasn't using the OCSS override. looks good

ashleygwilliams (approved):

PR #408, opened 2023-06-20
feat(vendor-css): vendor oranda-css by default

Gankra:
* remove oranda-css/dist/ from .gitignore so it can be checked in
* include_str! it, and use that if ORANDA_CSS and version aren't set
* add css upload to the post-release process
* change the specific-version-downloading logic to take a full tag


Gankra:
note: local oranda-css will only be updated *on cargo build*. the ORANDA_CSS var still lets you have it be updated more dynamically

manyinsects (changes requested):
This looks good code-wise, but there's two other things:

- All past oranda releases should have a corresponding CSS file attached so that this code works in a backwards compatible way
- It feels like we should have some system that guarantees that we always have a `main`-built CSS file checked in. Maybe a pre-commit hook?

ashleygwilliams (changes requested):
so generally speaking, we need to still publish the oranda.css and so i do not think we should check it in. i do think we should instead, build, vendor, and release it, as part of our release process. folks need the ability to pin the css version and so we need it to be requestable from our releases. it might be helpful to have a brief meeting around this - i can also try to draw something up

Gankra:
The purpose of checking it in is to enable contributors on main to always have a coherent view of the codebase without having to run `yarn`. The CI changes I made *will* get it uploaded to all future releases, and the format allows you to still ask for e.g. "css-v0.0.1" to get the old system. I don't think it's necessary to upload this to old versions of oranda releases, as no such legacy connection exists.

Gankra:
closing as per https://gist.github.com/Gankra/621dee83748efe2b6b3ce6b9dc6c0650

PR #419, opened 2023-06-21
fix: properly escape paths when generating urls

Gankra:
also make logo linking properly use link::generate

### DIFF
--- a/src/site/layout/header.rs
+++ b/src/site/layout/header.rs
@@ -9,12 +9,18 @@ use axohtml::{html, text};
 use indexmap::IndexMap;
 
 fn get_logo(logo: String, config: &Config) -> Result<Box<img<String>>> {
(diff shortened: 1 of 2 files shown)


Gankra:
Note that this still leaves a latent ~aesthetic issue of us not unescaping urls when writing them to disk. The problematic URL is 

https://cdn.wikirby.com/archive/0/04/20200203000421%21Kirby_JP_Twitter_Icon.png 

which should be saved/displayed as

`20200203000421!Kirby_JP_Twitter_Icon.png`

with this PR we will save/display it as

`20200203000421%21Kirby_JP_Twitter_Icon.png`

producing a URL of 

`https://mywebsite.com/20200203000421%2521Kirby_JP_Twitter_Icon.png`

(double-escaping it)

PR #430, opened 2023-06-22
feat(disable-artifacts): make artifacts more disableable

Gankra:
You can now set artifacts: false to completely disable it

You can now set artifacts.hidden = [tarball, zip] to hide those tabs


ashleygwilliams (approved):
this looks good- would be nice to add a test or two- but maybe worth opening a new issue for "testing new artifacts config" since a lot has changed and no new tests have been added (e.g. for preferred and additional as well)

PR #440, opened 2023-06-22
renable ignored tests

Gankra:


Gankra:
with #442 the deleted assert can be added back with the same text as its twin

ashleygwilliams:
closing in favor of #444

PR #442, opened 2023-06-22
fix: make the installer widget show up more

Gankra:


Gankra:
Screenshots of the result applied to opensource.axo.dev/oranda

(NOTE: THESE EXAMPLES ARE ARTIFICAL AND I AM NOT PRUNING THE DETAILS, SO YOU MAY SEE THINGS LIKE "tarball" AS A TAB WHEN THE DESCRIBED SCENARIO WOULDN'T SPECIFICALLY HAVE IT, BUT THE GENERAL LAYOUT IS STILL WHAT IT WOULD BE)


# Normal

image


# Empty Github Release

```rust
platforms_we_want.clear();
```

This occurs when The Artifacts Feature Is Enabled but with no package-managers (say with cargo-dist: true), they have Github Releases, but the "latest" one we smart-select has no assets (say because they haven't published with cargo-dist yet, but have existing Github Releases with just source tarballs). 

image


# One Platform (Incidental)

```rust
platforms_we_want = platforms_we_want.into_iter().take(1).collect::<HashMap<_, _>>();
```

This is a really exotic situation that's similar to the previous one but we find Exactly One single-platform thing in the release (a tarball for one platform), and some other universal stuff (package_managers) so we have concrete information about the platform of interest, and a few options to show for it.

image


# One Platform (Universal)

```rust
platforms_we_want = platforms_we_want.into_iter().take(1).map(|(k, v)| ("all".to_owned(), v)).collect::<HashMap<_, _>>();
```

This is the much more normal situation where we have stuff to show but it's all "universal" so there's no point in mentioning platforms. This is basically the default UX for someone *only* using package_managers but who has Github Releases to get versions/dates off of. 

(errata: such a situation wouldn't have "tarball" visible, as the Universal path is only taken if there are no "archives". It also wouldn't show "installer.sh" in this situation because there are package_managers, which are "more universal" than the unix-only installer.sh. installer.sh would however show up if it was the only option, because now the universe is constrained to unix. This fuckery exists to prevent us from freaking out and listing 24 platforms when you have a bunch of package managers and one semi-specific thing like installer.sh. In the future it would be Nice to have more rich platform grouping but this would still be a nightmare to display unambiguously.)

image


# One Universal Download

```rust
platforms_we_want = platforms_we_want.into_iter().take(1).map(|(k, mut v)| ("all".to_owned(), vec![v.pop().unwrap()])).collect::<HashMap<_, _>>();
```

This is One Platform (Universal) but the only thing to show is a single Download. This would only happen if we detect there's something to download but can't detect its platform. Normally we gate treating something as a download *on* the presence of target information, so if this is at all possible it's because of some codepath I didn't finish writing. It shouldn't happen normally.

But hey, neat that it works.

image


# One Universal Script

```rust
platforms_we_want = platforms_we_want.into_iter().take(1).map(|(k, mut v)| ("all".to_owned(), vec![v.remove(1)])).collect::<HashMap<_, _>>();
```

This is One Platform (Universal) but the only thing to show is a single Runnable Oneliner. This is the default for someone who sets exactly one package_manager and nothing else, or who just has a detect install.sh.

image


# Tarball City

```rust
    platforms_we_want = platforms_we_want.into_iter().map(|(k, v)| (k, v.into_iter().filter(|i| matches!(release.artifacts.installer(*i).method, InstallMethod::Download { file: _ })).collect::<Vec<_>>())).collect::<HashMap<_, _>>();
```

This is the situation when you have Github Releases full of platform-specific tarballs and nothing else. So cargo-dist with `installers = []`, `taiki-e/create-github-release`, etc.

image


# No Github Releases

Vanilla code, I just set `repository = "https://github.com/Gankra/sburb"` which is a Github Repo with no Github Releases. As such we fallback to ReleaseSource::Current which uses the version in our Cargo.toml (can be overriden with project.version) and shows no date. This is the default UX for someone using package_managers but who hasn't ever published a Github Release.

(errata: I had to point this at github because we error out failing to parse the URL as a Github Repo. That sucks for e.g. bitbucket users who want to still use package_managers. Unfortunately it's hard to both allow this and tell the user that setting their github repo gets them nice things. Maybe a warning? I filed #443)

image

Gankra:
Cases I forgot to enumerate:

* failed platform detection (they have linux and mac but you're on windows)
* noscript
* mobile
* too many tabs that need to linewrap (we don't handle this correctly, they just overflow the box; you can see this easily on mobile)

Gankra:
Oh also forgot artifacts: false

image

ashleygwilliams (approved):

PR #447, opened 2023-06-25
fix broken images on opensource.axo.dev

untitaker:
make all image links relative so that they work when deployed on
...axo.dev/oranda/


manyinsects (approved):
Thank you!

PR #452, opened 2023-06-26
feat: support toml syntax highlighting

manyinsects:
adds TOML syntax highlighting by way of https://github.com/jasonwilliams/sublime_toml_highlighting

- question: said highlighting syntax file is MIT-licensed, can we vendor this? i've included the license header in the actual file
- to dump the syntax set to disk, i've made a normally uncalled function that, to update the set in the future, can be inserted anywhere in the call path. i'm not super happy with this approach (would prefer to add a hidden command or something), but i figured it's the path of least resistance given the time constraint


manyinsects:
Closes #236

src/site/markdown/syntax_highlight/TOML.sublime-syntax
```diff
@@ -0,0 +1,429 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/3/syntax.html
+# MIT License
+#
+#Copyright (c) 2021 Jason Williams
```
ashleygwilliams, line 6: IANAL but based on my previous experience and understanding, i think this is the right way to handle this. oranda is licensed MIT/Apache-2.0 which is compatible

ashleygwilliams (approved):
let's talk in the meeting about this - i assume it's probably Fine, or at least Fine For Now, but my naïveté around syntax themes makes me less than confident on evaluating this approach without more context :)

Issue #426, opened 2023-06-22
Swap quickstart images in docs to new default theme


manyinsects:

manyinsects:
Make sure to update it again if there's any huge visual changes with the #394 stuff

PR #454, opened 2023-06-26
docs: update images and quickstart

manyinsects:
switched to new theme and widget images, and rewrite the quickstart guide a bit to use a repo that doesn't already have oranda configured (but does have cargo-dist)


manyinsects:
Closes #426

PR #469, opened 2023-06-27
fix: support subdirectories for additional pages paths

### LINKED ISSUE #448: How to link to an additional page from README?
I have a feature overview on https://untitaker.github.io/spacemod/ where I would like to link into the "Matching modes" page.

I would like my README to link to `./docs/matching.md` (so following the link in GitHub's UI works), while the oranda landing page should link to `https://untitaker.github.io/spacemod/matching/`.

Is that possible or should I go with mdbook? I find mdbook overkill for simple docs.

manyinsects:
```
{
  "build": {
    "additional_pages": {
      "Subpage": "subdir/mypage.md",
    }
  }
}
```

now builds and links to a page at `yoursiteurl.com/subdir/mypage`. Closes #448.

### DIFF
--- a/src/site/layout/header.rs
+++ b/src/site/layout/header.rs
@@ -53,11 +53,10 @@ fn nav(
         Message::new(MessageType::Info, "Found additional pages...").print();
         for (page_name, page_path) in additional_pages.iter() {
             if page::source::is_markdown(page_path) {
-                let file_path = page::source::get_filename(page_path);
+                let file_path = page::source::get_filename_with_dir(page_path);
 
                 if let Some(file_name) = file_path {
-                    let href =
-                        link::generate(path_prefix, &format!("{}/", file_name.to_string_lossy()));
+                    let href = link::generate(path_prefix, &format!("{}/", file_name));
 
                     html.extend(html!(<li><a href=href>{text!(page_name)}</a></li>));
                 } else {
--- a/src/site/layout/header.rs
+++ b/src/site/layout/header.rs
@@ -53,7 +53,7 @@ fn nav(
         Message::new(MessageType::Info, "Found additional pages...").print();
         for (page_name, page_path) in additional_pages.iter() {
             if page::source::is_markdown(page_path) {
-                let file_path = page::source::get_filename_with_dir(page_path);
+                let file_path = page::source::get_filename_with_dir(page_path)?;
 
                 if let Some(file_name) = file_path {
                     let href = link::generate(path_prefix, &format!("{}/", file_name));
--- a/src/site/page/mod.rs
+++ b/src/site/page/mod.rs
@@ -60,6 +60,22 @@ impl Page {
         })
     }
 
+    pub fn new_from_file_with_dir(source: &str, layout: &Layout, config: &Config) -> Result<Self> {
+        let body = Self::load_and_render_contents(source, &config.styles.syntax_theme)?;
+        let contents = layout.render(body, None);
+        // Try diffing with the execution directory in case the user has provided an absolute-ish
+        // path, in order to obtain the relative-to-dir path segment
+        let relpath = if let Some(path) = pathdiff::diff_paths(source, std::env::current_dir()?) {
+            path
+        } else {
+            source.into()
+        };
+        Ok(Page {
+            contents,
+            filename: relpath.display().to_string(),
+        })
+    }
+
     pub fn new_from_contents(
         body: String,
         filename: &str,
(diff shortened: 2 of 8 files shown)


src/site/layout/header.rs
```diff
@@ -53,11 +53,10 @@ fn nav(
         Message::new(MessageType::Info, "Found additional pages...").print();
         for (page_name, page_path) in additional_pages.iter() {
             if page::source::is_markdown(page_path) {
-                let file_path = page::source::get_filename(page_path);
+                let file_path = page::source::get_filename_with_dir(page_path);
 
                 if let Some(file_name) = file_path {
-                    let href =
-                        link::generate(path_prefix, &format!("{}/", file_name.to_string_lossy()));
+                    let href = link::generate(path_prefix, &format!("{}/", file_name));
```
ashleygwilliams, line 59: we should join paths with camino

src/site/page/mod.rs
```diff
@@ -60,6 +60,22 @@ impl Page {
         })
     }
 
+    pub fn new_from_file_with_dir(source: &str, layout: &Layout, config: &Config) -> Result<Self> {
```
ashleygwilliams, line 63: could just accept syntax theme instead of all config i think?

src/site/page/source.rs
```diff
@@ -12,3 +13,21 @@ pub fn get_filename(file: &str) -> Option<&OsStr> {
     let file_path = Path::new(file);
     file_path.file_stem()
 }
+
+pub fn get_filename_with_dir(file: &str) -> Option<Utf8PathBuf> {
```
ashleygwilliams: am slightly worried about error handling/unwraps here though that might be me be overbearing- what do you think?

src/site/page/source.rs
```diff
@@ -12,3 +13,21 @@ pub fn get_filename(file: &str) -> Option<&OsStr> {
...
+    // path, in order to obtain the relative-to-dir path segment
+    let path = if let Some(path) =
+        pathdiff::diff_utf8_paths(file, Utf8PathBuf::from_path_buf(cur_dir).unwrap())
+    {
+        path
+    } else {
+        Utf8PathBuf::from(file)
+    };
+
+    Some(path.with_extension(""))
```
ashleygwilliams: what exactly does this mean?

tests/build/fixtures/oranda_config.rs
```diff
@@ -8,10 +8,8 @@ use oranda::site::javascript::analytics::Plausible;
 
 pub fn no_artifacts(temp_dir: String) -> Config {
     let mut additional_pages = IndexMap::new();
-    additional_pages.insert(
-        "Another Page".to_string(),
-        "https://raw.githubusercontent.com/axodotdev/oranda/main/README.md".to_string(),
-    );
+    additional_pages.insert("Another Page".to_string(), "docs/src/cli.md".to_string());
```
ashleygwilliams, line 11: nice

ashleygwilliams (commented):
some small things, might just need to understand better

src/site/page/source.rs
manyinsects: this is basically `path.file_stem()` but it operates on the entire path instead of just the file name, so:

`Path::from("foo/bar/baz.txt").file_stem()` -> `baz`

`Path::from("foo/bar/baz.txt").with_extension("")` -> `foo/bar/baz`

src/site/page/mod.rs
manyinsects: i figured the signature should be similar to the function above (which it is based on)

src/site/layout/header.rs
manyinsects: there's no path joining being done here, unless you mean the URL path, which is handled by the `link::generate` function, the `format!` is just for handling the dev-only trailing-slash quirk that we've talked about before

src/site/page/source.rs
manyinsects: ye, will refactor to return a result

src/site/page/mod.rs
ashleygwilliams: ah word, we can leave this for a future pass on that then

src/site/page/source.rs
ashleygwilliams: oh interesting.

src/site/layout/header.rs
ashleygwilliams: ah yeah whoops, i forgot

ashleygwilliams (approved):
🙏

Issue #305, opened 2023-05-12
issues if cargo-dist enabled but haven't yet published with it


Gankra:
"Can't have artifacts header with no artifacts"

You can see this on https://github.com/Gankra/socc-pair

with 

```json
{
    "repository":"https://github.com/Gankra/socc-pair",
    "artifacts": { "cargo_dist": true }
}
```

We should just warn and move on, I think? This isn't the nicest behaviour for someone trying to bring everything up at once who just wants to try things out without running cargo-dist in prod.

Gankra:
This is no longer a crash, but with #382 this situation can be *auto-detected* and you get completely empty widgets/pages 
![image](https://github.com/axodotdev/oranda/assets/1136864/2c67a400-7b6a-4646-be50-52cb00a31d29)

Gankra:
This would be solved with #397.

Gankra:
This is essentially fixed now. You can get an empty "install" page with this precise config, but that's basically fine if you're going to publish Soon and you can force it off with cargo_dist: false.

Gankra:
Gonna reopen to signify decision on to what extent we should show empty UI to say "yep you enabled this, but we have no content"

Gankra:
We now show this in the empty case

![image](https://github.com/axodotdev/oranda/assets/1136864/0f9d5a5f-fce0-4502-845f-d21c572ff7d7)

Gankra:
criteria for for-reals closing this: make the widget and install page consistent

By doing exactly one of these two options:

* Making the Install Page hide when empty
* Making the Widget show when empty

ashleygwilliams:
should be finished, test to be sure

PR #470, opened 2023-06-27
fix(context): check if has github releases

### LINKED ISSUE #397: refine artifacts.has_some()
> blegh, auto-detecting release stuff has annoyingly inverted the responsibility of artifacts.has_some()
>
> because we don't try to build a Context if we don't has_some, but at this point has_some is kinda a question that's only answered by building a Context
>
>(ideally we would give you an "install" page if we find a Github Release with things that look like installables, based only on your Cargo.toml having repo defined)
>
> gonna leave it for now, bugfixable later (workaround is just to add any kind of package manager to your config)
>
> building contexts is very expensive and that's why it is useful to not do it eagerly
>
> at worst we can add artifacts: true

ashleygwilliams:
fixes #397 

do not merge this until #437 is merged as this is rebased on that; can review now tho

### DIFF
--- a/src/data/github/repo.rs
+++ b/src/data/github/repo.rs
@@ -1,3 +1,6 @@
+use std::fmt;
+
+use crate::data::GithubRelease;
 use crate::errors::*;
 
 use miette::{miette, IntoDiagnostic};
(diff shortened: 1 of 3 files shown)


ashleygwilliams:
will need to test for https://github.com/axodotdev/oranda/issues/305 to make sure this doesn't goof that

manyinsects (approved):

PR #474, opened 2023-06-27
feat: detect intel mac if there's no m1 mac artifacts

### LINKED ISSUE #456: suggest intel mac releases with M1 mac
if a release only has intel mac binaries and the user is running a m1 mac, show em the intel releases cuz they can run those too

manyinsects:
Closes #456


ashleygwilliams (commented):
i think if we decide to merge this we should display a note

ashleygwilliams (approved):

PR #487, opened 2023-06-28
fix: be more specific about having no releases with cargo-dist

### LINKED ISSUE #305: issues if cargo-dist enabled but haven't yet published with it

manyinsects:
Closes #305 for good


manyinsects:
previews:

image

image

PR #507, opened 2023-07-03
[WIP] Add install-widget to release pages

Gankra:
This uses the fact that the install-widget was always release-specific, so we can generate different copies for every install page

### DIFF
--- a/src/site/artifacts/mod.rs
+++ b/src/site/artifacts/mod.rs
@@ -66,11 +66,16 @@ pub fn scripts(release: &Release, config: &Config) -> Result<Vec<Box<div<String>
     Ok(output)
 }
 
-pub fn header(context: &Context, config: &Config) -> Result<String> {
+/// Build the install-widget for the latest release
+pub fn header(context: &Context, config: &Config) -> Result<Box<div<String>>> {
     let Some(release) = context.latest() else {
-        return Ok(String::new());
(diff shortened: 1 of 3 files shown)


Gankra:
Screen Shot 2023-07-03 at 2 22 51 PM

Gankra:
Also note that, as discussed in the Long Long Ago, "view source" here is not the Rich View Source we give to the latest release. It hotlinks the github download for the script, to avoid downloading+hosting a million scripts.

Gankra:
"View all installation options" will also link the downloads page, which is just the latest download. 🤷‍♀️

Gankra:
The package-manager strings are also "wrong" in that they aren't templated to specify the version to install.

manyinsects (approved):
I think this works as a "suboptimal" feature for now (I think it satisfies Most use-cases)

ashleygwilliams (changes requested):
i think there's design work we need to do before this is shipped- can talk more in the product meeting on thursday

manyinsects:
Closing due to several reasons:

- We most likely want to avoid making a changelog page _too_ much like the index page
- There would have to be more design decisions around templating (where do we insert the version to install? is it up to the user? what syntax does this use? etc)
- This PR is incompatible with the ongoing work around separating the install widget from oranda itself

PR #508, opened 2023-07-03
fix cupcake theme preview

untitaker:
may i interest you in https://github.com/untitaker/hyperlink/ to check the generated HTML for broken image urls? ;)


ashleygwilliams (approved):
thanks for this! and yes- i think we'd like a link checker!! exactly *how* we incorporate it might be a bit complicated but if you wanted to propose something we'd def be interested!

Gankra:
adding data-point: for mdbook stuff there is https://github.com/Michael-F-Bryan/mdbook-linkcheck


(I had this setup in cargo-dist's CI but currently have disabled because it was freaking out at the way oranda ran mdbook (i think fixed in #287 by making the book path absolute))

untitaker:
@Gankra i opened #509, let me know if you have a strong preference for mdbook-linkcheck

PR #509, opened 2023-07-03
add link checker to CI

untitaker:
see #508 

CI should fail, and then after rebase it shouldn't.


untitaker:
~it seems that PRs do not run the docs job, and I don't think that's intentional. my best guess is that this problem is limited to PRs from external contributors due to `permissions: contents: write` in web.yml~ I messed up the gha yml syntax/schema and gha silently skipped the workflow

manyinsects (approved):

PR #511, opened 2023-07-03
Add an arm64 apple fallback hack

Gankra:


Gankra:
note that this isn't tested, does anyone have a project that would trigger this?

manyinsects:
I have a test condition for this, will give it a try

manyinsects:
on aarch64 apple with a release only containing x86 linux/macos binaries: 

CleanShot 2023-07-04 at 12 40 15@2x

manyinsects (approved):

Issue #493, opened 2023-06-29
Workspace project support


jamesmunns:
Hi there! I chatted privately with @ashleygwilliams about this, but I'd be very interested in using oranda for [mnemos](https://github.com/tosc-rs/mnemos) to replace the, uh, [current state of the art](https://mnemos-dev.jamesmunns.com/) of our landing page.

Mnemos is a workspace, and has a number of "internal" (developer facing) and "external" (user facing) libraries and applications. Having some way to have a "collection" of pages/projects/??? would be very neat!

Not sure if this is already tracked somewhere, but i'd be interested in trying oranda out again if this is a use case that is supported in the future.

Thanks!

ashleygwilliams:
we were not yet tracking it anywhere but this is a feature we are all pretty excited about. we've heard the request in two forms, "workspace" and "org"- but the general idea is "multi-tenant" oranda. we can use this as a tracking issue and keep you up-to-date as we make progress :)

PR #512, opened 2023-07-03
WIP: rewrite relative links to additional pages

Gankra:
this is currently hardcoded to SECURITY.md, and needs a more robust notion of relative links to pages we care about

(also this temporarily adjusts the repo to give a test case)


ashleygwilliams:
from @Gankra: 

> this PR needs much more work, but i think a lot of that work will happen while we work on monorepo (#493) stuff
the missing piece is a mapping from page ~names to page ~paths, so that the link rewriter knows what to rewrite.
>
> that PR is basically providing the logic for editing the results of the markdown parser before rendering to html which I wanted to confirm was doable. ideally before we start rendering pages we have a listing of every page we will render so relative linking can work well (and maybe this also reduces the amount of need we have for path-prefix, if every page knows where exactly all other pages are in the hierarchy?)

PR #526, opened 2023-07-11
Refactor into using `minijinja` templates instead of `axohtml`

manyinsects:
Refactors the code to use minijinja instead of axohtml.


src/site/artifacts/mod.rs
```diff
@@ -1,76 +1,143 @@
...
-mod table;
+use crate::data::artifacts::inference::triple_to_display_name;
+use crate::message::{Message, MessageType};
+use crate::site::javascript;
+use serde::Serialize;
 
-use axohtml::elements::div;
-use axohtml::{html, text};
+type DownloadableFiles = Vec<(FileIdx, (File, Vec<Option<String>>))>;
+type Platforms = HashMap<TargetTriple, Vec<InstallerIdx>>;
```
Gankra, line 21: If this is getting fed into templates, do we want it to be BtreeMap/IndexMap? (It was a HashMap only because I knew it wouldn't show up in output)

Gankra: Also iirc I had some doc-comments describing how these are interpreted, might be worth transplanting them here too?

src/site/artifacts/mod.rs
```diff
@@ -1,76 +1,143 @@
...
+    let os_script = javascript::build_os_script_path(&config.build.path_prefix);
+    let platforms_with_downloads = filter_platforms(release);
+
+    let mut downloadable_files: Vec<_> = release
+        .artifacts
+        .installers()
+        .filter_map(|(_, installer)| {
+            let InstallMethod::Download { file } = installer.method else {
+            return None;
+        };
```
Gankra: wait did rustfmt actually generate this? 0.o

(i think rustfmt support for let-else might only be on the latest toolchain)

src/site/funding/mod.rs
```diff
@@ -1,155 +1,142 @@
...
-/// Generate the standalone funding page.
-pub fn page(config: &Config, funding: &Funding) -> Result<String> {
-    let mut funding_items = funding.content.clone();
-    // We've already made sure that we can unwrap on all of these `Option`s
+pub fn context(config: &Config, funding: &Funding) -> Result<FundingContext> {
+    let mut funding_base = funding.content.clone();
     let unwrapped_config = config.components.funding.as_ref().unwrap();
-    let preferred_html = if unwrapped_config.preferred_funding.is_some() {
+    let preferred_funding = if unwrapped_config.preferred_funding.is_some() {
         let preferred = unwrapped_config.preferred_funding.as_ref().unwrap();
```
Gankra: should be updated to an if-let

src/site/funding/mod.rs
```diff
@@ -1,155 +1,142 @@
...
-        // Remove the preferred item from the rest of the list
-        funding_items.remove(preferred);
-        preferred_funding_section(preferred.clone(), funding.content.clone())
+        let content = funding_base.get(preferred).cloned().unwrap();
+        funding_base.remove(preferred);
```
Gankra: Is this unwrap correct? What if the user typo's the key?

Regardless you can replace this with `let content = funding_base.remove(preferred);` (give or take the unwrap)

src/site/layout/header.rs
```diff
@@ -23,140 +19,12 @@ async fn fetch_logo(
     path_prefix: &Option<String>,
     dist_dir: &str,
     origin_path: String,
-    name: &String,
-) -> Result<Box<img<String>>> {
+    _name: &str,
+) -> Result<String> {
```
Gankra: Does some later code properly apply this now..? (maybe rename the function to get_logo_url if so..?)

src/site/mod.rs
```diff
@@ -232,14 +249,21 @@ impl Site {
                 &config.styles.syntax_theme,
             )?;
         }
+        if config.styles.favicon.is_some() {
+            let copy_result_future = Asset::copy(
+                config.styles.favicon.as_ref().unwrap(),
```
Gankra: should be if-let

src/site/oranda_theme.rs
```diff
@@ -1,28 +1,17 @@
...
+)]
 #[serde(rename_all = "lowercase")]
 pub enum OrandaTheme {
     Light,
     Dark,
-    #[serde(alias = "axo_light")]
+    #[serde(alias = "axo_light", rename(serialize = "axo"))]
     AxoLight,
-    #[serde(alias = "axo_dark")]
+    #[serde(alias = "axo_dark", rename(serialize = "dark axo"))]
```
Gankra, line 23: wait, the light theme is "the" axo theme? that's backwards, no?

src/site/templates.rs
```diff
@@ -0,0 +1,97 @@
...
+use crate::errors::Result;
+use crate::site::layout::LayoutContext;
+use crate::site::markdown::SyntaxTheme;
+use crate::site::{link, markdown};
+use include_dir::{include_dir, Dir};
+use minijinja::value::Value;
+use minijinja::{context, AutoEscape, Environment, Template};
+use std::collections::HashMap;
+
+const TEMPLATE_DIR: Dir<'_> = include_dir!("$CARGO_MANIFEST_DIR/templates");
```
Gankra, line 19: woah, wild that include_dir provides this sugar

src/site/templates.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+impl<'a> Templates<'a> {
+    pub fn new(config: &Config) -> Result<Self> {
+        let mut env = Environment::new();
+        let mut files = HashMap::new();
+        Self::load_files(&TEMPLATE_DIR, &mut files).unwrap();
+        for (path, contents) in files {
+            env.add_template_owned(path, contents).unwrap();
+        }
```
Gankra: these shouldn't ever fail in prod but it would be nice to make these expects to save debugging time when developing?

src/site/templates.rs
```diff
@@ -0,0 +1,97 @@
...
+        let template = self.env.get_template(name)?;
+        Ok(template.render(context_with_layout)?)
+    }
+
+    fn load_files(dir: &Dir, files: &mut HashMap<String, String>) -> Result<()> {
+        for entry in dir.entries() {
+            if let Some(file) = entry.as_file() {
+                let file_path = file.path();
+                // Remove the .j2 extension
+                let file_name = file_path.with_extension("");
```
Gankra, line 67: TIL this is an officially documented way to pop one part of the extension

src/site/templates.rs
```diff
@@ -0,0 +1,97 @@
...
+        };
+        link::generate(&path_prefix, &base)
+    }
+
+    fn syntax_highlight(code: String, lang: String, _syntax_theme: String) -> String {
+        // TODO: Fix when we support more syntax themes.
+        let syntax_theme = SyntaxTheme::MaterialTheme;
+        match markdown::syntax_highlight(Some(&lang), &code, &syntax_theme) {
+            Ok(res) => res,
+            Err(_) => format!("<code class='inline-code'>{code}</code>"),
```
Gankra, line 98: Is this like the last place in our code where we randomly have some inline-html in rust?

Gankra (approved):
A few nits about unwraps, but looks good. Only reviewed the rust, didn't review the templates (seems impractical to try to do).

Gankra:
Pushed up some resolutions of most of my comments. I ended up doing a quick pass over all the `unwraps` in non-test code and either replaced them or upgraded them to an `expect` with some description of why they shouldn't happen in prod, ostensibly.

I added a couple TODOs for genuine questions I don't want to unilaterally make a call on, that should be resolved before merging.

src/site/artifacts/mod.rs
```diff
@@ -1,76 +1,152 @@
...
+use serde::Serialize;
 
-use axohtml::elements::div;
-use axohtml::{html, text};
+/// A list of downloadable files.
+///
+/// The inner Vec is a list of supported platforms (display name). If the Option is None then
+/// we don't have a pretty name for the platform and will omit it. (This seems suboptimal,
+/// shouldn't we show the ugly TargetTriple name as fallback?)
+type DownloadableFiles = Vec<(FileIdx, (File, Vec<Option<String>>))>;
```
Gankra: TODO

src/site/artifacts/mod.rs
```diff
@@ -1,76 +1,152 @@
...
+/// A map from TargetTriples to Installers that support that platform
+///
+/// In theory this should be BTreeMap or IndexMap but something in the pipeline from here to
+/// jinja seems to be forcing a sorting so it's deterministic..? Can't find docs for this.
+type Platforms = HashMap<TargetTriple, Vec<InstallerIdx>>;
```
Gankra, line 21: SEMI-TODO

src/site/funding/mod.rs
```diff
@@ -1,155 +1,146 @@
...
+    let mut funding_base = funding.content.clone();
+    // allowed temporarily to preserve the TODO
+    #[allow(clippy::manual_map)]
+    let preferred_funding = if let Some(preferred) = config.preferred_funding.as_ref() {
+        if let Some(content) = funding_base.remove(preferred) {
+            Some(to_funding_methods(preferred, &content))
+        } else {
+            // TODO: check if we warn about this anywhere earlier, the user clearly messed up here
+            // and specified a preferred funding that doesn't exist, and should be told about it
+            None
```
Gankra: TODO

src/site/templates.rs
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    fn triple_to_display_name(target: String) -> String {
+        // TODO: should this be `unwrap_or(target)` to produce some useful output?
+        triple_to_display_name(&target).unwrap_or("").to_string()
+    }
```
Gankra: TODO

tests/build/mod.rs
```diff
@@ -1,11 +1,14 @@
...
-use oranda::site::layout::Layout;
 
 mod fixtures;
 use super::utils::tokio_utils::TEST_RUNTIME;
 use fixtures::{oranda_config, page};
 
+// TODO:
+// Fix these tests by using a library that's actually HTML-aware so that we don't have to mess around
+// with trying to detect variable levels of whitespace.
+
```
Gankra: Should be a FIXME instead?

Gankra (changes requested):

src/site/oranda_theme.rs
Gankra: still TODO

tests/build/mod.rs
Gankra: https://github.com/axodotdev/oranda/issues/528

Gankra (approved):

ashleygwilliams (approved):

PR #532, opened 2023-07-18
feat: add workspace capabilities

### LINKED ISSUE #493: Workspace project support

manyinsects:
## What's this?

This PR adds the capability for oranda to build multiple sites at once. For now, this needs to be set up manually. To do this, create a new `oranda-workspace.json` file. This file has the same syntax as the regular oranda configuration file, but by virtue of its name, oranda knows that it should be used to construct a workspace. The workspace file can be filled with these new keys:

```json
{
  "workspace": {
    "name": "My Workspace", // optional
    "members": [
      {
        "slug": "projectone",
        "path": "./project_one"
      },
      {
        "slug": "projecttwo",
        "path": "./project-two"
      }
    ]
  }
}
```

The output tree for this will look like this:

```ascii
├── public
│   ├── index.html
│   ├── oranda-v0.2.0-prerelease.1.css
│   ├── projectone
│   │   ├── artifacts.js
│   │   ├── index.html
│   │   └── oranda-v0.2.0-prerelease.1.css
│   └── projecttwo
│       ├── artifacts.js
│       ├── index.html
│       └── oranda-v0.2.0-prerelease.1.css
```

In `oranda-workspace.json`, you can also define all regular oranda config keys, which will be passed down to each member's config (with the member custom config taking precedence). Depending on what you pass down, this can lead to very weird behavior (such as if you pass down a list of package managers).

## TODO

- [x] Add a workspace index page that links between members
- [ ] Add a toggle to disable generating this page -> #534 
- [x] Find out why oranda css is also put in the workspace root `public` directory
- [ ] Refactor site generation to not have to rely on changing the directory, instead working off a given "root path"
- [x] Refine CLI output to prefix member site output lines with the slug of the site being built
- [x] Figure out if I implemented root path_prefix correctly?
- [ ] Potentially, maybe out-of-scope, depends on time: Integrate `axoproject` workspace detection -> #535 

## Misc

Closes #493.


src/site/mod.rs
```diff
@@ -105,7 +138,20 @@ impl Site {
...
+    }
+
+    pub fn get_workspace_config() -> Result<Option<Config>> {
+        let path = Utf8PathBuf::from("./oranda-workspace.json");
+        if path.exists() {
+            let workspace_config = Config::build_workspace_root(&path)?;
+            Ok(Some(workspace_config))
+        } else {
+            Ok(None)
+        }
```
Gankra, line 167: Probably will want to use https://docs.rs/axoproject/latest/axoproject/fn.find_file.html and/or https://docs.rs/axoproject/latest/axoproject/struct.WorkspaceInfo.html#structfield.workspace_dir in the longterm.

(Depends how hard we want `oranda build` to work if you're not in the root dir, being strict like this is an ok starting point)

src/site/mod.rs
```diff
@@ -125,7 +171,7 @@ impl Site {
 
     fn print_plan(config: &Config) {
         let mut planned_components = Vec::new();
-        if config.components.artifacts.is_some() {
+        if config.components.artifacts_enabled() {
```
Gankra, line 187: good catch!

src/config/mod.rs
```diff
@@ -135,6 +139,37 @@ impl Config {
...
+        config_path: &Utf8PathBuf,
+        root_config_path: &Utf8PathBuf,
+        project_root: &Utf8PathBuf,
+    ) -> Result<Config> {
+        let member_conf = OrandaLayer::load(config_path)?;
+        let root_conf = OrandaLayer::load(root_config_path)?;
+        let project = AxoprojectLayer::load(Some(project_root.into()))?;
+        let mut cfg = Config::default();
+        cfg.apply_custom_layer(root_conf);
+        cfg.apply_project_layer(project);
```
Gankra: This ordering is very interesting, did you have a case in mind where this works Better than axoproject->oranda-workspace.json->oranda.json?

Gankra: Also HMM it's unfortunate that axoproject is being invoked for every package, when it already computes info for every package in a cargo workspace unconditionally. But it's not as simple as running it only once, because we allow for flexible members that might not be properly in a cargo workspace. Feels like we want to have some kind of cache where we check if previous runs of axoproject picked up this dir, and only rerun if not...?

hmm hmm hmm I need to think about the API design for axoproject here

manyinsects: in my mind, the oranda workspace config should not be able to override whatever axoproject figures out

ashleygwilliams: hrrrrrm, i am not sure of the exact details of what you mean here, but i think in general explicit oranda config should *definitely* override all other configs, largely due to it being the only config within our and our users direct control. anything that axoproject suggests should be modifiable (with additions or substractions) by an oranda-config. it's possible this leads a user to errors, but they should be easier to fix than errors in axoproject.

manyinsects: >  explicit oranda config should definitely override all other configs

it does, you can ofc still override all things in the workspace member config, my first thought was just that axoproject should have priority over what's passed down from the workspace root

manyinsects: (happy to change this, ofc)

src/site/mod.rs
```diff
@@ -214,7 +276,12 @@ impl Site {
         Ok(())
     }
 
-    pub fn write(self, config: &Config) -> Result<()> {
+    pub fn write(self, config: Option<&Config>) -> Result<()> {
+        let config = if self.workspace_data.is_some() {
+            &self.workspace_data.as_ref().unwrap().config
+        } else {
+            config.unwrap()
+        };
```
Gankra: should definitely have at least one if-let here, not sure what the config.unwrap() is about

src/errors.rs
```diff
@@ -110,6 +110,12 @@ pub enum OrandaError {
         details: AxoassetError,
     },
 
+    #[error("Unable to create the correct paths we need to process your site.")]
+    #[diagnostic(help(
+        "It can help to have your workspace members in a subdirectory under your workspace root."
+    ))]
+    PathdiffError { root_path: String, path: String },
+
```
Gankra: unused fields?

src/config/axoproject.rs
```diff
@@ -146,11 +143,7 @@ impl AxoprojectLayer {
                 } else {
                     // Found a workspace but none of the packages specifically control this dir.
                     // This can happen if you run oranda in a dir with a virtual Cargo.toml.
-                    Message::new(
-                        MessageType::Warning,
-                        &format!("Ignoring {:?} project, oranda is currently per-package and this looks like a whole workspace", workspace.kind),
-                    )
-                    .print();
+                    tracing::warn!("Ignoring {:?} project, oranda is currently per-package and this looks like a whole workspace", workspace.kind);
                     None
```
Gankra, line 147: Should this still be the case? I guess so since you reload all of axoproject for every oranda.json?

src/config/axoproject.rs
```diff
@@ -166,14 +159,7 @@ impl AxoprojectLayer {
...
                 None
             }
-            axoproject::WorkspaceSearch::Missing(cause) => {
-                // Just quietly log this in case useful
-                tracing::info!(
-                    "Couldn't find a {name} project {:?}",
-                    &miette::Report::new(cause)
-                );
-                None
-            }
```
Gankra: Why no more logging of this?

src/config/mod.rs
```diff
@@ -135,6 +139,37 @@ impl Config {
...
+        cfg.apply_custom_layer(member_conf);
+        cfg.apply_autodetect_layer()?;
+        Ok(cfg)
```
Gankra: this ordering is still sus to me, but i'll think about it while i do the rest of the review

src/data/mod.rs
```diff
@@ -1,212 +1,225 @@
...
+    /// the artifacts subsystem.
+    pub has_artifacts: bool,
+}
+
+impl Context {
+    /// Make a Context with a faux-release for the current project state
+    pub fn new_current(
+        project_config: &ProjectConfig,
+        artifacts_config: Option<&ArtifactsConfig>,
+        prefix: &Option<String>,
```
Gankra: wait what's up with this prefix arg? it's not path_prefix, but like... a logging prefix..?

src/main.rs
```diff
@@ -57,6 +60,11 @@ fn run(cli: &axocli::CliApp<Cli>) -> Result<(), Report> {
         .build()
         .expect("Initializing tokio runtime failed");
     let _guard = runtime.enter();
+    let sub_filter = tracing_subscriber::filter::Targets::new().with_target("oranda", Level::DEBUG);
+    let sub = tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::layer().event_format(formatter::OrandaFormatter))
+        .with(sub_filter);
+    let _sub_guard = set_default(sub);
```
Gankra: is this fighting against axocli at all, or is this all happy..?

src/site/layout/css.rs
```diff
@@ -26,17 +25,18 @@ pub fn get_css_link(
...
 }
 
-fn fetch_css(dist_dir: &str, release_tag: &str) -> Result<String> {
+fn fetch_css(dist_dir: &str, release_tag: &str, prefix: &Option<String>) -> Result<String> {
     match env::var("ORANDA_CSS") {
         Ok(path) => {
             let filename = "oranda.css".to_string();
             let msg = format!("Overriding oranda_css path with {}", &path);
-            Message::new(MessageType::Warning, &msg).print();
+            tracing::warn!(prefix, "{}", &msg);
```
Gankra: increasingly feeling weird that some logging arg needs to be threaded everywhere...

src/site/layout/mod.rs
```diff
@@ -132,4 +133,25 @@ impl LayoutContext {
             social: config.marketing.social.clone(),
         })
     }
+
+    /// Generates a new layout context to use for the index page.
+    pub fn new_for_index(workspace_config: &Config) -> Result<Self> {
```
Gankra: this probably wants to have a name/doc that screams "WORKSPACE" a bit more loudly, since "the index page" is gonna be a really confusing term of art to use for this feature.

src/site/layout/mod.rs
```diff
@@ -132,4 +133,25 @@ impl LayoutContext {
...
+            &workspace_config.build.path_prefix,
+            &workspace_config.styles.oranda_css_version,
+            &None,
+        )?;
+        Ok(Self {
+            project_name: workspace_config
+                .workspace
+                .name
+                .clone()
+                .unwrap_or("Oranda Workspace".to_string()),
```
Gankra: This default should be set in the Default impl of WorkspaceConfig (also the md docs say it will be "My Oranda Workspace" iirc?).

src/commands/build.rs
```diff
@@ -1,46 +1,66 @@
...
+            config_path: config_path.unwrap_or(Utf8PathBuf::from("./oranda.json")),
+        }
+    }
+
+    pub fn run(&self) -> Result<()> {
+        if let Ok(Some(config)) = Site::get_workspace_config() {
+            let sites = Site::build_multi(&config)?;
+            // FIXME: Let the user turn this off
+            if true {
+                tracing::info!("Building workspace index page...");
```
Gankra, line 40: say more about this FIXME?

src/commands/build.rs
```diff
@@ -1,46 +1,66 @@
...
+                let mut member_data = Vec::new();
+                for site in &sites {
+                    member_data.push(site.workspace_data.clone().unwrap());
```
Gankra, line 46: why is this an unwrap?

src/site/link.rs
```diff
@@ -1,30 +1,99 @@
...
+/// The function will attempt to lazily build the smallest possible absolute and canonicalized path,
+/// before diffing it with the root path to create a path that's always relative to the workspace root.
+///
+/// Some example scenarios:
+/// 1. root path = "/my/directory", member path = None, path = "myfile.md"
+///    Output = "myfile.md"
+/// 2. root path = "/my/directory", member path = "member", path = "myfile.md"
+///    Output = "member/myfile.md"
+/// 3. root path= "/my/directory", member path = "/my/directory/member", path = "../root.md"
+///    Output = "root.md"
```
Gankra, line 47: greatly appreciate these examples!

src/formatter.rs
```diff
@@ -0,0 +1,110 @@
...
+                    arrow_icon,
+                    style.apply_to(message.message)
+                )
+            } else {
+                format!("TRACE: {}", message.message)
+            };
+
+            if message.prefix.is_some() {
+                write!(&mut writer, "[{}] {}", message.prefix.unwrap(), output_str)?;
+            } else {
```
Gankra: if-let

Gankra: Also if you move this up to the top of the function you can just directly `write!` the rest of the message instead of buffering it up in `output_str`

src/formatter.rs
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    fn record_debug(&mut self, field: &Field, value: &dyn Debug) {
+        if field.name() == "message" {
+            self.message = format!("{:?}", value);
+        }
+        if field.name() == "prefix" {
+            self.prefix = Some(format!("{:?}", value));
+        }
```
Gankra: a bit surprised to see you debug formatting the message/prefix, won't that add extra quotes..?

src/formatter.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/// Style of output we should produce
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, ValueEnum)]
+pub enum OutputFormat {
+    /// Human-readable output
+    Human,
+    /// Machine-readable JSON output
+    Json,
+}
```
Gankra, line 164: Is this getting used?

src/data/workspaces.rs
```diff
@@ -0,0 +1,64 @@
...
+        // Set the correct path prefix. This should be:
+        // - If no root path prefix: `slug`
+        // - If root path prefix: `path_prefix/slug`
+        config.build.path_prefix = if workspace_config.build.path_prefix.is_some() {
+            // FIXME: Doesn't account for trailing slashes right now
+            Some(format!(
+                "{}/{}",
+                workspace_config.build.path_prefix.as_ref().unwrap(),
+                &member.slug
+            ))
```
Gankra: if-let

src/site/mod.rs
```diff
@@ -25,47 +26,98 @@ pub mod mdbook;
...
+    pub pages: Vec<Page>,
 }
 
 impl Site {
-    pub fn build(config: &Config) -> Result<Site> {
+    pub fn build_multi(workspace_config: &Config) -> Result<Vec<Site>> {
+        tracing::info!("Workspace detected, gathering info...");
+        // We assume the root path is wherever oranda-workspace.json is located (current dir)
+        let root_path = Utf8PathBuf::from_path_buf(std::env::current_dir()?.canonicalize()?)
+            .unwrap_or(Utf8PathBuf::new());
```
Gankra: can be unwrap_or_default()

src/site/templates.rs
```diff
@@ -45,7 +45,23 @@ impl<'a> Templates<'a> {
         env.add_filter("triple_to_display_name", Self::triple_to_display_name);
         // Use opt-in autoescape
         env.set_auto_escape_callback(|_| AutoEscape::None);
-        let layout = LayoutContext::new(config)?;
+        let layout = LayoutContext::new(config, prefix)?;
+        Ok(Self { env, layout })
+    }
+
+    pub fn new_for_index(workspace_config: &Config) -> Result<Self> {
```
Gankra: Similar note of `index` being a confusing term-of-art

src/site/workspace_index.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub repository: Option<String>,
+    pub logo: Option<Utf8PathBuf>,
+}
+
+impl WorkspaceIndexContext {
+    pub fn new(members: &Vec<WorkspaceData>, workspace_config: &Config) -> Result<Self> {
+        let mut map = IndexMap::new();
+        for member in members {
+            let logo = if member.config.styles.logo.is_some() {
+                Some(Self::find_logo_path(member, workspace_config)?)
```
Gankra: Should be an if-let that passes the logo in

src/site/workspace_index.rs
```diff
@@ -0,0 +1,71 @@
...
+            map.insert(member.slug.clone(), context);
+        }
+
+        Ok(Self { members: map })
+    }
+
+    fn find_logo_path(member: &WorkspaceData, workspace_config: &Config) -> Result<Utf8PathBuf> {
+        let logo_url = member.config.styles.logo.clone().unwrap();
+        let root_path = Utf8PathBuf::from_path_buf(std::env::current_dir()?.canonicalize()?)
+            .unwrap_or(Utf8PathBuf::new());
```
Gankra: unwrap_or_default

src/site/workspace_index.rs
```diff
@@ -0,0 +1,71 @@
...
+        if logo_url.starts_with("http") {
+            // Lifted from axoasset. Expose it there?
+            let mut filename = url::Url::parse(&logo_url)
+                .unwrap()
+                .path()
+                .to_string()
+                .replace('/', "_");
```
Gankra: this should probably produce an error instead of a panic?

src/site/workspace_index.rs
```diff
@@ -0,0 +1,71 @@
...
+            filename.remove(0);
+            let mut path = Utf8PathBuf::from(
+                workspace_config
+                    .build
+                    .path_prefix
+                    .clone()
+                    .unwrap_or("".to_string()),
```
Gankra: .as_ref().unwrap_or_default() should work fine here

src/commands/dev.rs
```diff
@@ -46,71 +48,38 @@ pub struct Dev {
...
-        paths_to_watch.push(config.project.readme_path);
-        // Watch for the oranda config file
-        paths_to_watch.push(
-            self.config_path
-                .clone()
-                .unwrap_or(Utf8PathBuf::from("./oranda.json"))
-                .into(),
-        );
+        let root_path = Utf8PathBuf::from_path_buf(std::env::current_dir()?.canonicalize()?)
+            .unwrap_or(Utf8PathBuf::new());
```
Gankra: unwrap_or_default

src/commands/dev.rs
```diff
@@ -198,4 +157,99 @@ impl Dev {
...
+            root_path,
+            &member_path,
+            config.project.readme_path,
+        )?);
+
+        // Watch for the oranda config file
+        let cfg_file = self
+            .config_path
+            .clone()
+            .unwrap_or(Utf8PathBuf::from("./oranda.json"));
```
Gankra: unwrap_or_else

src/commands/dev.rs
```diff
@@ -198,4 +157,99 @@ impl Dev {
...
+
+        // Watch for additional pages, if we have any
+        if !config.build.additional_pages.is_empty() {
+            let mut additional_pages = config
+                .build
+                .additional_pages
+                .values()
+                .cloned()
+                .map(|p| determine_path(root_path, &member_path, p).unwrap())
+                .collect();
```
Gankra: is this unwrap avoidable? (or why is it guaranteed?)

Gankra (changes requested):
The general approach seems fine, but I'm not quite able to fit it all in my head yet.

Mostly did a lot of flagging of unwraps and suggested simplifications, as well as some concerns around unclear naming of functions/args.

src/config/mod.rs
Gankra: Having mulled on it -- yeah I'd like to have workspace override project. If the user explicitly tells us something that should be preferred, and they can always just, not put anything in the workspace config.

Gankra: Also is there anywhere we actually complain if the user sets workspace stuff in a plain oranda.json? (very likely i just missed it)

src/site/layout/css.rs
Gankra: I think you mentioned wanting to use tracing spans for this kind of thing? (If so, big agree)

src/main.rs
manyinsects: axocli sets the global subscriber unfortunately, for now this sets the "local" subscriber, but we should eventually provide an option to axocli to skip setting up a subscriber

src/errors.rs
manyinsects: hm? the only time this is called is with both fields filled out

src/config/axoproject.rs
manyinsects: i expect this will get removed once i get around to implementing axoproject support for this

src/config/axoproject.rs
manyinsects: good idea, i'll move it to a `debug` log

src/config/mod.rs
manyinsects: > Also is there anywhere we actually complain if the user sets workspace stuff in a plain oranda.json? (very likely i just missed it)

not at the moment, but i'll add it

src/commands/build.rs
manyinsects: ideally, i want there to be a `workspace.skip_index_page` bool setting that allows workspaces to be used as a just a generate-multiple-sites feature without anything tying them together. planning to implement that this week

src/formatter.rs
manyinsects: don't think so, i think it's super legacy-ish

src/errors.rs
Gankra: What I mean is the error messages defined here don't actually use the fields of PathdiffError

src/main.rs
Gankra: filed https://github.com/axodotdev/axocli/issues/6

src/errors.rs
manyinsects: oh right, i guess i thought these fields were surfaced automatically

Gankra (approved):

PR #533, opened 2023-07-19
release: 0.2.0

manyinsects:


CHANGELOG.md
```diff
@@ -1,27 +1,50 @@
...
 
 - **make artifact autodetect configurable - [Gankra]/[pr527]**
 
+  We now provide a new boolean key, `components.artifacts.auto`, that lets you explicitly
+  enable autodetection of artifacts. Previously, we would only enable this if you either
+  turned on `components.artifacts.cargo_dist`, or if you provided some package manager entries.
+  Since oranda _does_ also support gleaning artifacts even without `cargo-dist` support enabled,
+  we added this extra switch to let you toggle it without having to mess around with package managers.
+
+  This is a **breaking change**, as enabling `cargo-dist` support or specifying package managers does
```
ashleygwilliams, line 15: for breaking changes i think it's helpful to have really *clear* update instructions. so while i think all the information is here in this section, we may want to say something like "if you were previously relying on this autodetection and update to this version, you'll see that your artifacts are no longer displayed on your oranda page. to fix this, create an `oranda.json` if you don't already have on and add this to it:

```json
{
   "artifacts": {
       "auto": "true"
  }
}
```

CHANGELOG.md
```diff
@@ -1,27 +1,50 @@
...
 ### Features
 
 - **Typescript syntax highlighting support - [shadows-withal]/[pr525], [geelen]/[i513]**
 
+  You can now use the `ts`/`typescript` languages in code blocks! Hooray for types!
+
 - **Better logo positioning - [SaraVieira]/[pr524], [tertsdiepraam]/[i519]**
 
+  Logos set via the `styles.logo` option will now be properly centered/aligned in all themes,
+  and set to a maximum width so that a 1920x1080 logo won't break the website completely.
```
ashleygwilliams: we may want to be more specific than "break website completely"

CHANGELOG.md
```diff
@@ -1,27 +1,50 @@
...
+
 ### Fixes
 
 - **Hacker theme highlight color - [SaraVieira], [pr523]/[i522]** 
 
+  Selecting text in the Hacker theme now applies a nice, green, high contrast highlight background
```
ashleygwilliams, line 46: we may want to say what it previously did

manyinsects: i'm not actually sure lol

ashleygwilliams (approved):
looks good to me! 🚀

PR #552, opened 2023-07-27
Allow more repos

### LINKED ISSUE #443: it's ok for repository to be set to not-github
image

image

If we have repository set then we insist on it having to be a github repo and error out if not. The github path already gracefully degrades to Self::make_current_release if there *is* Github and no Github Releases, that path should be taken (with a warning?) if parsing fails as well.

Plecra:
Fixes #443. The repository's will now be ignored if we can't read it, and render will continue to work.

I'm not sure this is the right way to handle this case - When the user has indicated they want releases to be automatically discovered (at least, not `auto: false` :)) and we can't reach the repo, it's quite likely somethings misconfigured and they'd want alarms to start sounding so they can fix it.

Is there a way we could signal a soft error? It's something that should be urgently addressed (compared to warnings) but we can limp on from for now.

### DIFF
--- a/src/site/mod.rs
+++ b/src/site/mod.rs
@@ -95,16 +95,18 @@ impl Site {
         Self::print_plan(config);
 
         if needs_context {
-            let mut context = match &config.project.repository {
-                Some(repo_url) => Context::new_github(
+            let mut context = config.project.repository.as_ref().and_then(|repo_url| {
+                Context::new_github(
                     repo_url,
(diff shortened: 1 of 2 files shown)


manyinsects:
This approach seems good to me, we basically already signal a "soft error" (which is what I'd call a warning), but it's something we can still refine at some later point.

manyinsects (approved):

PR #565, opened 2023-08-02
feat: use tailwind binary to build CSS

manyinsects:
Removes the dependency on a Node toolchain + Yarn to build CSS, as long as you're not running a release build. If you're running a release build, nothing changes, we still fetch the associated (or custom) CSS version from GitHub.

Also adds a new hidden command to generate _just_ the CSS for the purposes of attaching it to a release, and modifies the corresponding workflow.


SaraVieira:
The hero we needed

oranda-css/tailwind.config.js
```diff
@@ -52,7 +52,6 @@ module.exports = {
   plugins: [
     require("@tailwindcss/typography"),
     require("@tailwindcss/forms"),
-    require("@tailwindcss/line-clamp"),
```
manyinsects, line 55: (this is a default plugin in tailwind now)

src/site/layout/css.rs
```diff
@@ -28,27 +35,97 @@ fn concat_minify(css_files: &[String]) -> Result<String> {
...
     Ok(crate::site::link::generate(path_prefix, &filename))
 }
 
-fn fetch_css(dist_dir: &str, release_tag: &str) -> Result<String> {
+/// Places the CSS into the output folder. Depending on if we're running in local development or in
+/// a released binary, the method of obtaining said CSS will differ.
+pub fn place_css(dist_dir: &str, release_tag: &str) -> Result<()> {
+    // Even if you're running a development build, we still respect the custom CSS version preference
+    // by falling back to fetching said version from GitHub.
+    if cfg!(debug_assertions) && release_tag == ORANDA_CSS_TAG {
```
manyinsects, line 48: to my knowledge there isn't a better way to find out whether we're running a dev build or a release build, right?

src/site/layout/css.rs
```diff
@@ -28,27 +35,99 @@ fn concat_minify(css_files: &[String]) -> Result<String> {
...
+    LocalAsset::create_dir_all(&binary_path)?;
+    binary_path.push(format!("tailwindcss-{double}"));
+    if !binary_path.exists() {
+        // Fetch the binary from GitHub if it doesn't exist
+        tracing::info!("Fetching Tailwind binary from GitHub release...");
+        let url = format!(
+            "https://github.com/tailwindlabs/tailwindcss/releases/latest/download/tailwindcss-{double}"
+        );
+        let handle = tokio::runtime::Handle::current();
+        let response = handle.block_on(reqwest::get(url))?;
```
manyinsects, line 85: FYI: `axoasset` currently doesn't handle requests for a binary correctly, so for now I'm falling back to using `reqwest` (which is already in our dependency tree)

src/site/layout/css.rs
Gankra: It's either this or `cfg(test)` (neither of which would work for my habit of developing with --release). This is fine enough.

src/site/layout/css.rs
Gankra: Oh because it insists everything is utf8 text?

src/site/layout/css.rs
```diff
@@ -28,27 +35,99 @@ fn concat_minify(css_files: &[String]) -> Result<String> {
...
+        ("linux", "x86_64") => "linux-x64",
+        ("linux", "aarch64") => "linux-arm64",
+        ("linux", "arm") => "linux-armv7",
+        ("macos", "x86_64") => "macos-x64",
+        ("macos", "aarch64") => "macos-arm64",
+        ("windows", "x86_64") => "windows-x64.exe",
+        ("windows", "aarch64") => "windows-arm64.exe",
+        _ => "linux-x64",
+    };
+    let mut binary_path = Utf8PathBuf::from(cache_dir.display().to_string());
```
Gankra, line 75:
```suggestion
    let mut binary_path = Utf8PathBuf::from_path(cache_dir)?;
```

src/site/layout/css.rs
manyinsects: no, because it does very naive mime type checking which basically only accounts for images, so it gets fussy and quits

src/site/layout/css.rs
```diff
@@ -28,27 +35,99 @@ fn concat_minify(css_files: &[String]) -> Result<String> {
...
+                .parent()
+                .expect("Tailwind binary path has no parent!?"),
+        )?;
+
+        // On non-Windows platforms, we need to mark the file as executable
+        if !double.starts_with("windows") {
+            Command::new("chmod")
+                .args(["+x", binary_path.as_str()])
+                .output()?;
+        }
```
Gankra:
```suggestion
        #[cfg(target_family = "unix")]
        {
            use std::os::unix::prelude::PermissionsExt;
            let user_execute = std::fs::Permissions::from_mode(0o755);
            std::fs::set_permissions(binary_path, user_execute)?;
        }
```

I think?

src/site/layout/css.rs
```diff
@@ -28,27 +35,99 @@ fn concat_minify(css_files: &[String]) -> Result<String> {
...
+        .args([
+            "-c",
+            "oranda-css/tailwind.config.js",
+            "-i",
+            css_src_path.as_str(),
+            "-o",
+            &format!("{dist_dir}/oranda.css"),
+            "--minify",
+        ])
+        .output()?;
```
Gankra, line 115: Might also want to check `output.status.success()`. Command returning an Err is generally an OS error like "binary_path wasn't executable at all".

Gankra (approved):
Seeeems reasonable?

src/site/layout/css.rs
manyinsects: doesn't work, `Utf8PathBuf` only has `from_path_buf`, which is annoyingly a side-effectful operation

manyinsects: (`Utf8Path` has `from_path`, but we need an owned path)

PR #566, opened 2023-08-02
feat: inline css on release

### LINKED ISSUE #554: Improve CSS loading
1. We should replace our current Yarn setup with on-demand compiling with a [Tailwind binary](https://github.com/tailwindlabs/tailwindcss/releases/tag/v3.3.3)
2. We should by default include the resulting CSS in our release builds, while still letting the user override CSS if they specify a different CSS version

manyinsects:
Closes #554. On release, this will pre-build the CSS and drop it into `oranda-css/dist/oranda.css`, as well as set an environment variable letting oranda know that it can inline that version.


Cargo.toml
```diff
@@ -50,12 +53,15 @@ pathdiff = { version = "0.2.1", features = ["camino"] }
 minijinja = { version = "1.0.3", features = ["loader", "preserve_order"] }
 include_dir = "0.7.3"
 slug = "0.1.4"
-directories = "5.0.1"
+oranda-generate-css = { path = "oranda-css" }
```
Gankra: any reason to have the crate not have the same name as the dir..?

Gankra: Also you need to have an actual version here or `cargo publish` will fail, as published crates can only depend on crates in crates.io. In general making this a crate means we now need to publish this to crates.io as its own thing.

The PublishRelease ci task will also need to be updated to `cargo publish ...` this package before publishing oranda.

Also versioning this in lockstep with oranda itself will streamline release processes.

(I possibly led you astray with the note of ".crates are just a tarball of the whole dir" because I'm *pretty sure* that the obvious implications of that re: path dependencies doesn't work -- cargo/crates.io will still complain on publish.)

build.rs
```diff
@@ -0,0 +1,7 @@
+fn main() {
+    // Build the CSS on-demand if we're running a release-ish build here (as determined by Cargo)
+    if std::env::var("PROFILE").is_ok_and(|v| v == "release") {
+        oranda_generate_css::build_css("oranda-css/dist/").unwrap();
```
Gankra: Will this work right if you run `cargo build` from a weird dir?

src/site/layout/css.rs
```diff
@@ -43,84 +38,27 @@ pub fn get_css_link(path_prefix: &Option<String>, release_tag: &str) -> Result<S
...
+        Ok(oranda_generate_css::build_css(dist_dir)?)
+    } else if maybe_env.is_some() && release_tag == ORANDA_CSS_TAG {
+        // If we're in a release build, we can inline a file generated by oranda's build script.
+        // Don't even build this if we're not in a release build so that Clippy et al don't complain
+        #[cfg(not(debug_assertions))]
         {
-            use std::os::unix::prelude::PermissionsExt;
-            let user_execute = std::fs::Permissions::from_mode(0o755);
-            std::fs::set_permissions(&binary_path, user_execute)?;
+            let css = include_str!("../../../oranda-css/dist/oranda.css");
```
Gankra, line 53: Is this not a compiler error for non-release builds? I would have thought all include_strs are resolved statically, so the fact that it's in a branch doesn't matter...

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8441ab4ff8c4f663a7c39d8ed947da3f

Oh I see you wrapped this in `#[cfg(not(debug_assertions))]`... hmmm... I guess that should work? But at that point you might as well write the code as:

```rust
if release_tag == ORANDA_CSS_TAG {
    #[cfg(debug_assertions)]
    {
        // build css
    }
    #[cfg(not(debug_assertions)]
    {
        // write static css
        include_str!("../../oranda-css/dist/oranda.css")
    }
} else {
    // fetch css
}
```

And just not have the env var at all.

Gankra (changes requested):
I expect this won't work when you try to cargo-publish (hopefully --dry-run will complain appropriately)

Cargo.toml
manyinsects: > any reason to have the crate not have the same name as the dir..?

mainly because the directory also contains the actual oranda css, and it feels like signalling "oh this only contains rust code" if we name it that way. but we could also separate the directories properly

src/site/layout/css.rs
manyinsects: > But at that point you might as well write the code as: <snip>

this makes sense to my sleep-addled brain, i'm pretty sure there was some objection i had to this path a couple hours earlier, i'll see if i remember tomorrow

Cargo.toml
manyinsects: good call on the version, that's kind of annoying but makes sense from Cargo's perspective

build.rs
manyinsects: yeah, `build.rs` runs from the cwd of the Cargo.toml, it seems

Cargo.toml
manyinsects: okay, `oranda-generate-css` is published, i'll set the correct perms tomorrow

src/commands/print.rs
```diff
@@ -19,8 +19,8 @@ pub struct GenerateCss {}
 
 impl GenerateCss {
     pub fn run(&self) -> Result<()> {
-        oranda::site::layout::css::build_css("dist")?;
-        tracing::info!("CSS placed in dist/oranda.css");
+        oranda_generate_css::build_css("oranda-css/dist")?;
```
Gankra: this shows up in enough random places that it should probably be some kind of const in oranda_generate_css...

src/errors.rs
```diff
@@ -39,6 +39,9 @@ pub enum OrandaError {
     #[error(transparent)]
     UrlParse(#[from] url::ParseError),
 
+    #[error(transparent)]
+    GenerateCss(#[from] oranda_generate_css::errors::GenerateCssError),
+
```
Gankra:
```suggestion
    #[error(transparent)]
    #[diagnostic(transparent)]
    GenerateCss(#[from] oranda_generate_css::errors::GenerateCssError),

```

Lets miette know that the inner error is also a miette error, prints inner details right.

src/site/layout/css.rs
```diff
@@ -46,81 +41,24 @@ pub fn place_css(dist_dir: &str, release_tag: &str) -> Result<()> {
     // Even if you're running a development build, we still respect the custom CSS version preference
     // by falling back to fetching said version from GitHub.
     if cfg!(debug_assertions) && release_tag == ORANDA_CSS_TAG {
```
Gankra:
```suggestion
    if release_tag == ORANDA_CSS_TAG {
```

(forgot to remove this)

generate-css/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+        #[cfg(target_family = "unix")]
+        {
+            use std::os::unix::prelude::PermissionsExt;
+            let user_execute = std::fs::Permissions::from_mode(0o755);
+            std::fs::set_permissions(&binary_path, user_execute)?;
+        }
+    }
+
+    tracing::info!("Building oranda CSS using Tailwind...");
+    let css_src_path = Utf8PathBuf::from(CSS_SRC_PATH);
```
Gankra: You should make this relative to CARGO_MANIFEST_DIR (or something), otherwise `set_current_dir` in a test will break this (I've been running tests in --release while messing with gallery stuff to work around it).

.github/workflows/publish-release.yml
```diff
@@ -62,6 +62,9 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
+      - run: cargo publish -p oranda-generate-css --token ${CRATES_TOKEN}
+        env:
+          CRATES_TOKEN: ${{ secrets.CRATES_TOKEN }}
       - run: cargo publish --token ${CRATES_TOKEN}
```
Gankra: not 100% confident this will do the right thing, but it probably should...? (paranoid brain wants to add -p oranda here to be unambiguous)

Gankra (approved):
Seems like the right approach, some fixes suggested.

.github/workflows/publish-release.yml
manyinsects: yeah let's just do that to be sure

src/commands/print.rs
manyinsects: putting it in there would need us to publish a new version right now, which would break lockstep with oranda itself... kinda annoying

PR #567, opened 2023-08-02
Docs: update web.yml

mistydemeo:
The sample `web.yml` used for deploying points to a prerelease version of oranda; it should probably be using the latest instead. I also updated oranda's own `web.yml` to use the latest release.


mistydemeo:
Rolled back the Oranda upgrade in the workflow; it tripped the link checker.

ashleygwilliams (approved):

mistydemeo:
Updated to use the dynamic `latest` URL instead of 0.2.0.

PR #568, opened 2023-08-03
fix: canonicalize paths less

Gankra:
current_dir() is already absolute, and canonicalizing it on windows results in later path appends going wonky.


Gankra:
(There's still a few of canonicalize_utf8s in the code that seem more intentional (notably link.rs) that I'm leaving alone)

manyinsects (approved):

PR #570, opened 2023-08-03
Refactor dark theme + extras

### LINKED ISSUE #390: Improve dark mode handling in oranda-css
From Discord:

fwiw i think the way we're doing dark toggling is a bit cumbersome/brittle? Instead of having 

```css
.dark .blah {
  color: var(--dark-fg-color);
} 
.blah {
  color: var(--light-fg-color);
}
```

we can just do

```css
:root {
  --fg-color: var(--light-fg-color);
}
:root.dark {
  --fg-color: var(--dark-fg-color);
}

.blah {
  color: var(--fg-color);
}
```
no more "oops you forgot to add a dark variant for that component"

manyinsects:
Closes #390.

- Simplifies our handling of dark mode CSS by inverting variables
- Writes CSS for the workspace index page (unrelated bug)
- Fixes a nasty bug where changelog content would overflow if the content was too wide. Before and after:

CleanShot 2023-08-03 at 15 33 15@2x

CleanShot 2023-08-03 at 15 33 33@2x

### DIFF
--- a/oranda-css/css/pages/artifacts.css
+++ b/oranda-css/css/pages/artifacts.css
@@ -12,21 +12,17 @@
 
 .artifacts {
   @apply sm:flex items-center flex-col mb-8 p-0 hidden;
-  color: var(--light-highlight-fg-color);
-  background-color: var(--light-highlight-bg-color);
+  color: var(--highlight-fg-color);
+  background-color: var(--highlight-bg-color);
 }
 
 .artifacts-table {
   @apply block max-w-full overflow-auto;
 }
 
-.dark .artifacts {
-  color: var(--dark-highlight-fg-color);
-  background-color: var(--dark-highlight-bg-color);
-}
-
 ul.tabs {
   @apply flex border-b-2;
+  border-color: var(--highlight-fg-color);
 }
 
 ul.tabs li {
@@ -34,13 +30,8 @@ ul.tabs li {
 }
 
 ul.tabs li.selected {
-  color: var(--light-highlight-bg-color);
-  background-color: var(--light-highlight-fg-color);
-}
-
-.dark ul.tabs li.selected {
-  color: var(--dark-highlight-bg-color);
-  background-color: var(--dark-highlight-fg-color);
+  color: var(--highlight-bg-color);
+  background-color: var(--highlight-fg-color);
 }
 
 .install-content {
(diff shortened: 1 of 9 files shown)


oranda-css/css/pages/artifacts.css
```diff
@@ -12,35 +12,26 @@
...
 }
 
-.dark .artifacts {
-  color: var(--dark-highlight-fg-color);
-  background-color: var(--dark-highlight-bg-color);
-}
-
 ul.tabs {
   @apply flex border-b-2;
+  border-color: var(--highlight-fg-color);
```
Gankra, line 25: don't see an equivalent change this is mirroring, what's up..?

oranda-css/css/pages/changelog.css
```diff
@@ -37,29 +37,21 @@
...
 
 .releases-list {
   @apply flex flex-col gap-32;
 }
 
 .releases-wrapper {
-  @apply grid gap-12 relative items-start mt-12;
-  grid-template-columns: 160px 1fr;
+  @apply grid gap-12 relative mt-12;
+  grid-template-columns: 160px minmax(0, 1fr);
```
Gankra, line 49: another driveby fix..?

oranda-css/css/variables.css
```diff
@@ -1,10 +1,31 @@
...
+    --light-fg-color: #141414;
+    --light-bg-color: var(--dark-fg-color);
+    --dark-bg-color: var(--light-fg-color);
+    --fg-color: var(--light-fg-color);
+    --bg-color: var(--light-bg-color);
+
+    /* Link colors */
+    --light-link-color: #0284c7;
+    --dark-link-color: #8BB9FE;
+    --link-color: var(--light-link-color);
```
Gankra, line 13: any reason some of these are hardcoded here, and others are imported from themes/axo.css..?

Gankra (approved):
Is this not going to break all the other themes? Don't they need their vars renamed too?

oranda-css/css/pages/artifacts.css
manyinsects: this fixes an implied default border color that was looking pretty bad in axolight

oranda-css/css/pages/changelog.css
manyinsects: this is this: "Fixes a nasty bug where changelog content would overflow if the content was too wide."

oranda-css/css/variables.css
manyinsects: these colors are for the default theme (which is not our axo theme)

PR #575, opened 2023-08-03
Gallery demo

Gankra:
How to build the gallery:

```
cargo test gal_
```

How to serve the gallery:

```
cd target/tmp
cargo run -- serve
```

* [x] add harness
* [x] add build-pass tests
* [x] add workspace build-pass test ("the gallery")
* [x] snaphot testing of public/
  * [x] implement (currently filtered to all files that end in .html)
  * [x] fix: stable ordering of platform html in install widget (i think the `select` is always sorted but the actual entries the select reveals aren't)
  * [x] fix: filter the name of oranda.css (`cargo test` and `cargo test --release` write different file names)
* [x] fix: prevent hanging of gal_workspace (tries to wait for all other tests to signal success, so hangs if any fail to build)
* [x] fix: make `cargo test` work without `--release` (need to make the runtime generate-css function remember where CARGO_MANIFEST_DIR is instead of just going relative to cwd, so that the resulting binary can be run in different dirs on the same system)
* [x] feat: add a way to do multiple variants of the same repo that works with `oranda-workspace.json` (cargo-dist can happily edit the same checkout over and over, but oranda-workspace wants each variant to exist at the same time, so I'll probably just add a variant of `run_test` that takes a custom id (checkout dir name) so that you can checkout the same repo multiple times)
* [x] fix: ??? some way to prevent new Github Releases adding new data to the tests and causing snapshots to invalidate?
* [ ] fix: ??? visually distinguish variant gallery entries (if you have multiple copies of cargo-dist, workspace-index.html makes no visual distinction between the two, you can only tell based on the url of the "website" link)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
+      - uses: swatinem/rust-cache@v1
       - name: Run cargo test
         run: |
           cargo test --workspace
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,6 +1077,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6c80984affa11d98d1b88b66ac8853f143217b399d3c74116778ff8fdb4ed2e"
 
 [[package]]
+name = "glob"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d2fabcfbdc87f4758337ca535fb41a6d701b65693ce38287d856d1674551ec9b"
+
+[[package]]
 name = "globset"
 version = "0.4.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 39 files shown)


Gankra:
all the issues with oranda-gallery are fixed, but i realized a new one: snapshotting is fundamentally brittle to the fact that we source Github Releases for data, so any new Github Releases will change several pages and cause the tests to fail

(and having oranda-itself in the gallery means cutting an oranda release will break oranda's own tests... rip) 

Possible solutions:

* Don't snapshot test
* Incredibly intelligent snapshot test filtering
* Have a way to archive/clamp github release data to a specific date

**EDIT: implemented the 3rd one with DEBUG_DATA_CLAMP_DATE**

Gankra:
also a note on commit naming here:

* DEMO: things that are contained to the gallery itself
* DEMO-HACK: changes to oranda, several of which are genuine fixes that can/should be split off and merged in their own PRs (some may be genuine hacks tho)

Gankra:
Seems like a few system-specific slashes are making it into some actual html 😬:

```
 6724  6724 │ <p>These previews are generated by running oranda on itself.</p>
 6725  6725 │ <h2 id="dark-default"><a class="header" href="#dark-default">Dark (default)</a></h2>
 6726       │-<p><img src="configuration%5Ctheme/../../images/themes/dark.png" alt="dark theme preview" /></p>
       6726 │+<p><img src="configuration/theme/../../images/themes/dark.png" alt="dark theme preview" /></p>
```

manyinsects:
> Seems like a few system-specific slashes are making it into some actual html 😬

in mdbook output files, so this seems like a mdbook bug more than anything, we don't do any weird linking or anything afaik

manyinsects:
stopped mdbook html files from being output into snapshots, we don't want them there anyways, that's for mdbook itself to test

manyinsects (approved):

PR #579, opened 2023-08-09
feat: add markdown docs to workspaces

### LINKED ISSUE #564: workspaces: Provide markdown embed and preferred project(s)
Much like in the funding component, workspaces should be able to load content from a Markdown file, as well as have a notion to express that one member is the "main" or "preferred" member.

mistydemeo:
This enables embedding markdown documents into the workspace landing page, similar to how the funding page works today.

Refs #564.

### DIFF
--- a/templates/workspace_index/index.html.j2
+++ b/templates/workspace_index/index.html.j2
@@ -1,6 +1,10 @@
 {% extends "workspace_index/layout.html" %}
 
 {% block content %}
+{% if page.docs_content %}
+    {{ page.docs_content|safe }}
+{% endif %}
+
 <ul class="index-grid">
     {% for member in page.members %}
         <li>
--- a/src/config/workspace.rs
+++ b/src/config/workspace.rs
@@ -15,6 +15,8 @@ pub struct WorkspaceLayer {
     pub members: Option<Vec<WorkspaceMember>>,
     /// Whether to enable workspace autodetection
     pub auto: Option<bool>,
+    /// The path to additional documentation to render
+    pub docs_path: Option<String>,
 }
 
 #[derive(Debug, Serialize, Deserialize, JsonSchema, Clone, Hash, PartialEq, Eq)]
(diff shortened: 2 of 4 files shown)


templates/workspace_index/index.html.j2
```diff
@@ -24,4 +24,8 @@
         </li>
     {% endfor %}
 </ul>
-{% endblock %}
\ No newline at end of file
+
+{% if page.docs_content %}
+    {{ page.docs_content|safe }}
```
mistydemeo: I'm assuming the `safe` filter is fine here because the content is totally under the repo owner's control.

templates/workspace_index/index.html.j2
```diff
@@ -24,4 +24,8 @@
         </li>
     {% endfor %}
 </ul>
-{% endblock %}
\ No newline at end of file
+
+{% if page.docs_content %}
```
manyinsects: I believe we want this to sit above the workspace members (with the preferred members, people can still specify which members to put _above_ the content)

templates/workspace_index/index.html.j2
manyinsects: yep, should be

manyinsects (commented):
looking good! just one nit about content placement

templates/workspace_index/index.html.j2
mistydemeo: 👍 OK, pushed a change to put that above the members.

manyinsects (approved):

templates/workspace_index/index.html.j2
```diff
@@ -2,6 +2,10 @@
 
 {% block content %}
 <ul class="index-grid">
+{% if page.docs_content %}
```
manyinsects, line 4: oh, this needs to be outside the index grid ul (that's why i force pushed, sorry)

mistydemeo: I missed that you force-pushed! I force-pushed because I had to rebase on main, and didn't realize there had been changes. Made that change again

PR #580, opened 2023-08-09
feat: add "preferred" member to workspaces

mistydemeo:
This adds a context of a "preferred" workspace member, akin to the preferred funding methods. These preferred members can be displayed distinctly from the other members in a more prominent position. The "preferred" field is an array of strings, like so:

```json
{
  "workspace": {
    "preferred_members": ["member"],
  }
}
```

This needs a design pass from an HTML/visual perspective - I've got the feature plumbed through from the config, but the display is just a placeholder at this point.

Refs #564.


mistydemeo:
This version allows for multiple preferred members, but I can pare that back to a single preferred member if that's preferred.

manyinsects (approved):

Issue #584, opened 2023-08-16
Write artifacts context to separate JSON file


manyinsects:
In order to use oranda's artifacts inference stuff for integrations, we should write an `artifacts.json` file (toggleable in the config).

PR #589, opened 2023-08-16
feat: write json file with artifacts context

manyinsects:
Serializes the artifacts context to JSON and writes it to disk, under `<dist_dir>/artifacts.json`.

Lingering things

- [ ] ~Do we want to stick with this path? Do we want to make it customizable?~ - maybe later
- [ ] ~Do we want to give an option to disable generating this?~ - maybe later
- [x] Do we want an oranda flag that _only_ generates this file (and other relevant artifacts files like the installer txt files) - yes, done
- [x] Are we happy with this simplistic version keeping (`format_version`, in integrations you'd check if your version > format version) - we track oranda versions instead


manyinsects:
Closes #584

src/site/artifacts/mod.rs
```diff
@@ -15,14 +17,25 @@ use serde::Serialize;
 type DownloadableFiles = Vec<(FileIdx, File, Vec<String>)>;
 /// A map from TargetTriples to Installers that support that platform
 type Platforms = BTreeMap<TargetTriple, Vec<InstallerIdx>>;
-#[derive(Serialize, Debug)]
+
+/// Current version of the JSON artifacts format
+static JSON_VERSION: usize = 1;
```
Gankra: is it expected for a user to have to deal with mixed versions of this? i.e. will they ever have archived json blobs from previous releases, or will they all get rebuilt together?

PR #591, opened 2023-08-16
chore: use our own global to emulate set_current_dir in gallery

Gankra:
This prevents races with other tests, as we never change the current_dir of the actual test process, only the processes that it spawns.


mistydemeo (approved):
Line endings aside, looks good to me.

PR #595, opened 2023-08-17
release: 0.3.0

manyinsects:


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@
   **Migration instructions from 0.2.0 to 0.3.0**
 
   Set `components.changelog` to `false` in your `oranda.json` file if you previously didn't use the changelog feature.
+  Also, the changelog is now read from your _local_ rather than your _remote_ changelog file, and this is not a behavior that we
```
ashleygwilliams: there is config to change this tho right?

ashleygwilliams: i would use a variation/restatement of "Additionally, oranda will now attempt to read changelog information from a local `CHANGELOG(.md)` or `RELEASES(.md)`
  file, should it exist. If you want oranda to use GitHub release bodies instead, set
  `components.changelog.read_changelog_file = false`." instead of saying we don't want to restore behavior (that doesn't sound good)

PR #602, opened 2023-08-22
chore: enable Homebrew formula and tap

mistydemeo:
This makes use of the new Homebrew installer in [cargo-dist 0.2.0-prerelease.4](https://github.com/axodotdev/cargo-dist/releases/tag/v0.2.0-prerelease.4), including pushing the formula to our tap. I've used the prerelease cargo-dist to regenerate our release CI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,6 +50,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       has-releases: ${{ steps.create-release.outputs.has-releases }}
+      releases: ${{ steps.create-release.outputs.releases }}
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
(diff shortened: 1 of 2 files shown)


manyinsects (approved):
I'm assuming this was autogenerated by `cargo-dist`? if so, LGTM

.github/workflows/release.yml
```diff
@@ -115,7 +113,7 @@ jobs:
         with:
           submodules: recursive
       - name: Install Rust
-        run: rustup update "stable" --no-self-update && rustup default "stable""#
+        run: rustup update "stable" --no-self-update && rustup default "stable"
```
Gankra: we should listen to the cargo-dist warning and remove `rust-toolchain-version` from our Cargo.toml

mistydemeo: Good call, done.

PR #609, opened 2023-08-25
Fix the tailwind dependency breaking the Nix flake

hawkw:
Currently the `oranda-generate-css` crate will always attempt to
download a precompiled `tailwindcss` binary from GitHub Releases, and
store it in the user's cache directory. This doesn't work on platforms
such as NixOS, where the global `.cache` directory is not writable while
building a package. Thus, the Nix flake in the Oranda repo doesn't
actually work at all.

This branch changes `oranda-generate-css` to first check if a binary
named `tailwindcss` is on the current `PATH`, and uses that rather than
downloading the precompiled binary if Tailwind is already available. We
check for the existence of `tailwindcss` by running `tailwindcss --help`
and testing if the exit code is 0, which seemed like the most
cross-platform way to check if the binary exists. Potentially, we may
want to change this to check for a specific `tailwindcss` version in the
future, but I noticed that the current code downloads the latest
release, rather than a fixed version, so I wanted the same behavior
here.

Now that the `oranda-generate-css` crate will use a pre-existing
`tailwindcss` binary, we can modify the Nix flake to depend on the
[`tailwindcss` package from nixpkgs][1]. This avoids the need to write
to the cache directory, which is not writable inside of a Nix flake, as
well as the precompiled binary, which may not work on NixOS at all.

I noticed that the tailwindlabs/tailwindcss repo also contains a [Rust
implementation of the Tailwind CLI][2]. We may, eventually, want to
consider changing the approach used here to simply depending on the Rust
tailwind implementation as a library dependency, so that it can be
resolved by Cargo instead of downloaded from GitHub. However, that's a
larger change, and I wanted to do the simplest change that would fix the
Nix flake. Additionally, it might be desirable to add some kind of CI
step that ensures the flake builds successfully, if the maintainers want
to officially support it. That way, we could avoid breaking the flake in
the future.

[1]: https://search.nixos.org/packages?channel=unstable&show=tailwindcss&from=0&size=50&sort=relevance&type=packages&query=tailwindcss
[2]: https://github.com/tailwindlabs/tailwindcss/blob/master/oxide/crates/cli/src/main.rs


ashleygwilliams:
@hawkw quick q about how flakes work- is it building from source or is it able to point at a binary? the tailwind binary is really only necessary for local development. we "trigger" the download on a `cargo build` but there's no way to differentiate `cargo install`'s `cargo build` from a straight `cargo build` (as far as i know). our standalone binaries neither use nor download the tailwind binary- i am assuming that a flake is somehow running a `cargo install`?

hawkw:
@ashleygwilliams Nix builds everything from source. I'm not totally sure OTTOMH whether `rustPlatform.buildRustPackage` performs a `cargo build` or a `cargo install`, but I could check. In either case, it's necessary for the flake to build Oranda from source to distribute Oranda from NixOS.

In addition, the flake also provides a dev shell for local development, so it's not just for installing Oranda; it also provides an environment for Nix users to develop Oranda.

ashleygwilliams:
@hawkw thank you for elaborating, especially the last point. we've been talking internally about switching away from `build.rs` to a env var- though if Nix is also proving a development env for working on oranda- then it seems that tailwind would still be a dep (we'd ask users who are working on oranda locally to set the env var). is there a way to specify local dev dependences specifically? or would that just be a "dependency"

manyinsects:
(we may also go back to having a `package.json` so that distro packagers have an easier time specifying a well-known dependency rather than `tailwindcss`, which may not be in every registry. in any case, we'll allow both to coexist)

hawkw:
> is there a way to specify local dev dependences specifically? or would that just be a "dependency"

yes, flakes have a separate concept of "package(s)" and "dev shell(s)"; the "package" builds the actual packaged application, while a dev shell is typically the package's dependencies plus additional tools that are useful for development but not necessarily required to build the final artifact.

for example, the Oranda flake defines `packages.oranda` (and a default package which is the `oranda` package): https://github.com/axodotdev/oranda/blob/05039c315e94045c72daefe587564af1bcc2c15c/flake.nix#L76-L79
as well as a `default` dev shell which inherits the oranda package's dependencies: https://github.com/axodotdev/oranda/blob/05039c315e94045c72daefe587564af1bcc2c15c/flake.nix#L81-L109

the actual package and its dependencies are defined here: https://github.com/axodotdev/oranda/blob/05039c315e94045c72daefe587564af1bcc2c15c/flake.nix#L33-L68

manyinsects (approved):
This looks good to me, I'm hoping to push a change this week that'll remove the implicit Tailwind dependency (keeping it as an option behind a flag, basically), but this shouldn't break this, and would at most require a small change to include said flag. Thanks for taking the effort to do this!

PR #626, opened 2023-09-04
fix: refine bool layer application

### LINKED ISSUE #615: Merging workspace and member configuration doesn't prioritize member objects
From #611:

@jamesmunns:
> Also, the first thing I tried was to put "changelog": false in the workspace, then ONLY enabling the changelog w/ "pull from github" for our overview project, but I think that didn't work (no changelog generated for the overview project).

manyinsects:
Closes #615. I documented the function with all possible cases, too, but I need to check whether this change in behaviour (which I think is Correct) breaks any current ways we recommend people configure oranda.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -328,26 +328,42 @@ pub trait ApplyBoolLayerExt {
 
 impl<T> ApplyBoolLayerExt for Option<T>
 where
-    T: ApplyLayer,
+    T: ApplyLayer + Default,
 {
     type Inner = T::Layer;
+
+    /// Apply a layer that can either be a boolean, or a Layer Value (most likely an object).
+    ///
+    /// Possible cases (lhs is the resultant config, rhs is the incoming layer):
+    /// lhs == Some && rhs == true  = nothing happens
+    /// lhs == Some && rhs == false = lhs gets set to None
+    /// lhs == Some && rhs == value = layer gets applied to lhs
+    /// lhs == None && rhs == true  = lhs gets set to layer default
+    /// lhs == None && rhs == false = nothing happens
+    /// lhs == None && rhs == value = lhs gets set to layer default with layer applied
+    /// rhs = nothing               = we do nothing
     fn apply_bool_layer(&mut self, layer: Option<BoolOr<Self::Inner>>) {
         match layer {
             Some(BoolOr::Val(val)) => {
                 if let Some(this) = self {
                     this.apply_layer(val);
                 } else {
-                    // If self is None, then a previous layer completely disabled this.
-                    // For now we respect that and drop this request completely.
+                    let mut t = T::default();
+                    t.apply_layer(val);
+                    *self = Some(t);
                 }
             }
             Some(BoolOr::Bool(false)) => {
                 // Disable this setting
                 *self = None;
             }
-            Some(BoolOr::Bool(true)) | None => {
-                // Do nothing, no opinion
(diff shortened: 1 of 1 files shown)


Gankra (changes requested):
The way we do inference (mdbook and funding at least) will break with this change, as inference layers currently rely on `self = None` as a way to disable the functionality completely:

https://github.com/axodotdev/oranda/blob/e64fc3122e7ded68aadb1e60bfd5eae49817beca/src/config/components/mdbooks.rs#L100-L110

https://github.com/axodotdev/oranda/blob/e64fc3122e7ded68aadb1e60bfd5eae49817beca/src/config/components/funding.rs#L91-L100

manyinsects:
> The way we do inference (mdbook and funding at least) will break with this change, as inference layers currently rely on self = None as a way to disable the functionality completely:

Correct me if I'm wrong, but inference/autodetect gets applied _last_ and therefore is not affected by these changes, no? It basically always goes last and therefore even if it sets itself to `None`, it shouldn't get overridden by anything calling `apply_bool_layer` after it.

manyinsects:
Case in point: 
image

Gankra:
mm true, I might be thinking too abstractly (wanting previous layers to always properly work with theoretical future layers)

manyinsects:
fair, I guess it's weighing that against what users expect to be correct behaviour...

Gankra:
With this change does setting `mdbook = false` in the workspace config actually work? I would think that it would get clobbered by the package's own config defaulting to not-false.

Gankra:
We can maybe handle that by making `Some(true)` and `None` have different behaviours. Previously the user had no "inheritance" so there was no reason for "no opinion" and "true" to be different, but now explicitly setting = true to override false in the workspace is a thing.

manyinsects:
> With this change does setting mdbook = false in the workspace config actually work? 

it does:
image

manyinsects:
oh you meant the workspace config, one sec

manyinsects:
also works:
image

Gankra:
The case I'm specifically concerned about:

* workspace sets mdbook = false
* package wants to inherit that, doesn't mention mdbook at all
* oranda reads this as the package setting = true and clobbering the workspace

manyinsects:
this doesn't seem to happen, in my example it would have resulted the other dozen members all saying "Building components: mdbook". it seems like we're using the right behaviour here (auto-detect doesn't even happen because we explicitly auto-disabled it, this gets cascaded down correctly)

Gankra (approved):

Gankra:
Just for posterity: we hashed out the miscommunication in chat, and indeed the initial impl didn't merge workspace setting "= false" into package properly (package would clobber it). The fix was to make `= true` and `= None` genuinely different, with "true" being an explicit override and "None" being "don't change anything".

Gankra:
Also for posterity, with this change we now need to be cognizant of there being like... 3 "stages" of layering:

1. yolo adding whatever info we can (default layer, axoproject layer)
2. user explicit config that should win any disputes (oranda.json, oranda-workspace.json)
3. delicate/careful analysis (autodetect)
  a. auto-removing/validating info
  b. auto-adding info ONLY IF `lhs != None`
  
As long as we maintain this precise ordering of stages, things will work well. But if we start trying to mix them up, things will break (things getting turned back on after validation removed it, user settings getting overridden by axoproject, etc).

Issue #639, opened 2023-09-29
oranda build: "failed to read asset from" No such file or directory (os error 2)


konstin:
I tried `oranda build` (0.4.1) as shown on the website and the book on https://github.com/konstin/gourgeist 28c8126fad9b465b4ba6bd9170721594b6a04ed3:
```console
$ oranda build
↪ >o_o< INFO: Building components: artifacts, changelog
  × failed to read asset from
  ╰─▶ No such file or directory (os error 2)
  help: Make sure your path is relative to your oranda config or project manifest file.
```
I'm missing the information about what failed because it doesn't show the missing path

ashleygwilliams:
hey @konstin! this is a terrible error message- sorry about that! let me dig in and see what's up

manyinsects:
heya! sorry for that terrible error - that was an oversight with how we handle path fallbacks. I'll send a PR fixing that error to actually display the missing filename, which in your case was the readme file. one current "hard" requirement of oranda is that you kind of need a README file, this stems from the very first version, when literally all it did was render README files. oranda can do way more now, so I don't think a readme should be a requirement anymore.

manyinsects:
Posted https://github.com/axodotdev/oranda/pull/640, this should allow you to use oranda without a README present, and have much nicer error messages for all other paths that it expects to exist (mostly when you tell it explicitly that something should be at a specific path).

manyinsects:
I just released a new prerelease that contains the above PR, feel free to test it out, and let us know if you have any further issues or suggestions! https://github.com/axodotdev/oranda/releases/tag/v0.5.0-prerelease.1

konstin:
That fixes it, thank you!

PR #640, opened 2023-09-29
feat: relax file requirements, improve IO errors

manyinsects:
Behaviour changes:

- Readme file doesn't need to exist anymore, is now optional, with a warning being output
- oranda now hard errors when a hardcoded mdbook path can't be found
- local logo paths will now error when they're not found
- additional pages paths will also error when they're not found

TODO:

- [x] Add backticks around the path in the error message for readability
- [x] Update docs to mention that you don't need a readme circa 0.5.0

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -212,7 +212,7 @@ pub enum OrandaError {
 
     #[error("Specified path `{path}` was not found on your filesystem!")]
     #[diagnostic(
-        help = "Make sure you're in the same directory where your oranda.json or README file is!"
+        help = "Make sure you specify your path relative to the oranda.json/manifest file/README file of your project!"
     )]
     PathDoesNotExist { path: String },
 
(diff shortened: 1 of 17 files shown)


manyinsects:
Screenshots:

- Testing with the repo from #639 
image

- With a nonexistent hardcoded mdbook directory
image

src/errors.rs
```diff
@@ -210,6 +210,12 @@ pub enum OrandaError {
     #[diagnostic(help = "You can manually specify path in your components.mdbook config")]
     MdBookConfigInvalid,
 
+    #[error("Specified path `{path}` was not found on your filesystem!")]
+    #[diagnostic(
+        help = "Make sure you're in the same directory where your oranda.json or README file is!"
```
ashleygwilliams: we may want this to say "Make sure you specify your path relative to the README/oranda.json" though this is tricky because sometimes the error is due to an inferred path, where other times its a specified path. ideally- those 2 cases have different errors because they have different fixes

PR #652, opened 2023-10-03
Package specific releases

06chaynes:
This adds a new optional field to the artifacts config `package_specific_releases` which will default to false. When false things will work as they currently do, when true an additional check is enabled in `Context::with_releases` that will skip the release should the version tag not contain the project name.

This also modifies `Context::with_releases` to require a reference to ProjectConfig as an argument in order to perform the check against the project name.

### DIFF
--- a/src/config/components/artifacts/mod.rs
+++ b/src/config/components/artifacts/mod.rs
@@ -38,6 +38,7 @@ enum ArtifactSystem {
 pub struct ArtifactsConfig {
     pub auto: bool,
     pub cargo_dist: bool,
+    pub package_specific_releases: bool,
     pub package_managers: PackageManagersConfig,
     pub hidden: Vec<String>,
 }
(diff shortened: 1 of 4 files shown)


06chaynes:
Not sure why it's showing that entire file as changed atm, will review.

Edit: After adding the changes to `src/data/mod.rs` cargo fmt wants to do this apparently, at least when I run it.

Edit 2: Not really sure why that was happening but I redid the changes and now its happy without needing to change the entire file, so yay!

src/config/components/artifacts/mod.rs
```diff
@@ -38,6 +38,7 @@ enum ArtifactSystem {
 pub struct ArtifactsConfig {
     pub auto: bool,
     pub cargo_dist: bool,
+    pub package_specific_releases: bool,
```
manyinsects: While this is a good addition, I would change the name, maybe to `match_package_names` (seems more immediately obvious to me what it does that way). Also adding it to the docs reference page would be nice, but no worries, I can also do that.

06chaynes: I wasn't sure about the naming either, I'll get those updates made a bit later today. Thanks for the quick response!

06chaynes: Just pushed these changes, also added a doc entry but feel free to tweaks as needed!

06chaynes: actually ill fix that clippy

manyinsects (approved):
Perfect! Thanks!

PR #655, opened 2023-10-05
sets color on divs in chapter list items with filter

### LINKED ISSUE #654: draft mdBook chapter titles are unreadable in `oranda-dark` and `oranda-light` mdBook theme
mdBook has a concept of "draft chapters", where the chapter name appears in the table of contents but no page actually exists for that chapter. these chapter titles are greyed out in the table of contents. when rendering a mdBook using Oranda, the default `oranda-dark` and `oranda-light` themes color the the names of these draft chapters the same as the background color, rendering them unreadable. 

with all other mdBook themes provided by Oranda, the draft chapter titles are basically readable. for example, with the mdBook "navy" theme, they look like this:
image

with the `oranda-dark` theme, though, the draft chapter titles are the same as the background:
image

and similarly, with the `oranda-light` theme, they are white (as is the background):
image

06chaynes:
Adds:

```css
.chapter li.chapter-item div {
    color: var(--sidebar-fg);
    filter: brightness(60%);
}
```

to `oranda-css/mdbook-theme/css/chrome.css`

Fixes #654 because currently the div in the list item is inheriting from:

```css
.chapter li {
    display: flex;
    color: var(--sidebar-non-existant);
}
```

Produces:
Screenshot 2023-10-05 at 3 22 57 PM

### DIFF
--- a/oranda-css/mdbook-theme/css/chrome.css
+++ b/oranda-css/mdbook-theme/css/chrome.css
@@ -462,6 +462,11 @@ ul#searchresults span.teaser em {
     color: var(--sidebar-fg);
 }
 
+.chapter li.chapter-item div {
+    color: var(--sidebar-fg);
+    filter: brightness(60%);
+}
+
 .chapter li a:hover {
     color: var(--sidebar-active);
 }


manyinsects (approved):

manyinsects:
Thanks so much for all of these extremely good PRs! Is there any way we can send you some stickers or something to show some appreciation for your work?

06chaynes:
> Thanks so much for all of these extremely good PRs! Is there any way we can send you some stickers or something to show some appreciation for your work?

Thank you, means a lot! I'm more than happy to contribute to a project when and where I find I am able to do so. 

Looking forward to the next release!

PR #659, opened 2023-10-26
chore: cargo-dist 0.4.0

mistydemeo:
We can upgrade now that 0.4.0 is out.

Do we want to make a Linux static musl build?


manyinsects (approved):
I don't see why we shouldn't make a musl build?

PR #666, opened 2023-11-25
Add app name to label

### LINKED ISSUE #665: Improve displaying of multi-artifact releases
I have a project that consists of multiple binaries that belong to the same application and aren't ever released separately, so I push them all into the same release.

But the `cargo-dist` integration makes it look a little awkward

image

It would be great if the tabs could maybe have the application name in brackets (for example, `tar (my-project-cli)`).  
This is technically possible to derive from the `dist-manifest.json` that `cargo-dist` produces since it has the field `releases` which contains an array of objects that contain the `app_name` field.

aumetra:
Closes #665 

This is just a small proof-of-concept. To display the names, it just checks if there are more than 1 releases inside the `cargo-dist` metadata.  
If there is, it adds them in a smaller font to the next line.

image

This iteration still uses inline styles and some HTML inside the `write!` call which is not really *optimal*, I feel like.  
But I just opened this draft PR to showcase how it *could* look like.

### DIFF
--- a/src/data/cargo_dist.rs
+++ b/src/data/cargo_dist.rs
@@ -1,4 +1,6 @@
 use camino::Utf8PathBuf;
+use std::fmt::Write;
+
 pub use cargo_dist_schema::{ArtifactKind, DistManifest};
 
 use super::artifacts::{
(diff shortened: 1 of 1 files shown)


manyinsects:
Thanks for the proof of concept! I implemented this properly over at #671, happy to hear your thoughts :)

PR #667, opened 2023-11-29
feat: use extra artifacts instead of append release

mistydemeo:
The next version of cargo-dist includes a new feature that allows building and uploading extra artifacts as a part of the main build, without any `AppendRelease` trickery: https://github.com/axodotdev/cargo-dist/pull/613 To make the config a bit easier, I added a flag to `config-schema` to allow writing it to a file without piping.

This can't be merged until that new release of cargo-dist is out and we're updated to use it, however.


.github/workflows/additional-release-assets.yml
```diff
@@ -1,87 +0,0 @@
...
-      - name: Install oranda and build CSS
-        run: |
-          echo "uploading CSS to ${{ github.event.workflow_run.head_branch }}"
-          curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/oranda/releases/download/${{ github.event.workflow_run.head_branch }}/oranda-installer.sh | sh
-          oranda generate-css
-          ls oranda-css/dist
-          echo "built css!"
-          # Upload css to the Github Release™
-          gh release upload ${{ github.event.workflow_run.head_branch }} oranda-css/dist/oranda.css
-          echo "uploaded css!"
```
Gankra, line 87: it doesn't look like you added generation for oranda-css, and this is the Really Important one to migrate.

Gankra (changes requested):

.github/workflows/additional-release-assets.yml
mistydemeo: I actually totally missed that one - I'll add it now.

mistydemeo: Added.

Gankra (approved):

Gankra (approved):

PR #671, opened 2023-12-06
feat: show app names on multi-app releases

### LINKED ISSUE #665: Improve displaying of multi-artifact releases

manyinsects:
Currently only supports `cargo-dist`, and not filename inference. The names don't show if there's only one application included in the release.

image

Supersedes #666 and closes #665.


aumetra (approved):
That looks really good! Thanks for implementing that properly

PR #672, opened 2023-12-11
Support axo Releases

manyinsects:
Adds support for axo Releases (currently in closed beta) as a release data source for oranda. You can toggle between GitHub or axo Releases like this:

```json
{
  "components": {
    "source": "axodotdev"
  }
}
```

oranda will still default to GitHub, so no changes needed if you don't use axo Releases.

[gazenot]: https://github.com/axodotdev/gazenot


manyinsects:
(for easier reviewing, I broke the commit that homogenizes line endings out of the other work, the complete diff view is unreadable, but if you go commit by commit, it should be OK)

oranda-css/css/pages/changelog.css
```diff
@@ -3,6 +3,7 @@
 }
 
 .release-body {
+  margin-top: 2rem;
```
Gankra, line 6: ?

Gankra (approved):
looks right

oranda-css/css/pages/changelog.css
manyinsects: drive-by fix for the changelog title and body being too close

PR #678, opened 2023-12-19
fix: don't respect cargo-dist's install hints for curl-sh exprs

Gankra:
The dist-manifest.json fundamentally lacks the information to generate perfect pretty curl-sh exprs, because it needs to be uploaded before the create-release endpoint tells us this url format. As such, oranda should ignore the manifest for these install-hints, and instead use the inference backend for them, which can use the URLs returned by the list-releases endpoint (or the github release's base url).

### DIFF
--- a/src/data/cargo_dist.rs
+++ b/src/data/cargo_dist.rs
@@ -61,11 +61,14 @@ impl ReleaseArtifacts {
                             {
                                 preference = InstallerPreference::Custom;
                                 None
+                            } else if id.ends_with(".sh") || id.ends_with(".ps1") {
+                                // We have more info to do a better job than cargo-dist on
+                                // `curl | sh` exprs, inference will handle this for us!
+                                continue;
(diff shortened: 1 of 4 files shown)


Gankra:
NB when generating Github Release Bodies cargo-dist refines the dist-manifest with the info from create-release endpoint, so even cargo-dist knows that what it uploaded is subpar.

The only way for it to upload the right thing is for create-artifact-set to return a "hint" for what the URL would be if we did make a release.

...but that still wouldn't be totally correct for an install-widget for a hypothetical Release Preview, so really we should just stop considering cargo-dist authoritative on the Best Value for these exprs.

manyinsects (approved):

manyinsects:
@Gankra would you say this is worthy of a patch release?

Gankra:
yes, I don't want axo releases to be used with the current behaviour

PR #680, opened 2023-12-26
Update hosting.md

jovillarrealm:
To let it just reference earlier oranda.json and chapter 8, since this is the first time they would be edited but are not named.


manyinsects (approved):
Thanks!

Issue #683, opened 2024-01-27
Documentation incorrectly lists --project-root and --config-path parameters for oranda build 


lynnpepin:
Loving Oranda!! I just now found an issue where the docs refer to a feature that seems to not yet have been implemented. I wasn't able to find contributing guidelines, so I apologize if this issue isn't formatted correctly.

**TLDR:** [The Oranda docs](https://opensource.axo.dev/oranda/book/cli/build.html) suggests one can pass a parameter `--project-root` to `oranda build`, but one can't.

As of `0.6.1`, this does not work, and `oranda build -h` does not show the `--project-root` or `--config-path` parameters:

```
oranda build --help
Build an oranda site

Usage: oranda build [OPTIONS]

Options:
      --json-only
          Only build the artifacts JSON file (if applicable) and other files that may be used to support
          it, such as installer source files

  -h, --help
          Print help (see a summary with '-h')

GLOBAL OPTIONS:
  -v, --verbose
          Whether to output more detailed debug information

      --output-format <OUTPUT_FORMAT>
          The format of the output

          [default: human]

          Possible values:
          - human: Human-readable output
          - json:  Machine-readable JSON output
          
```

Digging into [the argument-parser source code](https://github.com/axodotdev/oranda/blob/main/src/commands/build.rs) shows a bit of context around the `build` command. 

> ``` 
>  /// DO NOT USE: Path to the root dir of the project
>  ///
>  /// This flag exists for internal testing. It is incorrectly implemented for actual
>  /// end-users and will make you very confused and sad.
> ```

(I've only tried `--project-root` but it seems `--config-path` has the same issues.)

lynnpepin:
My primary use case is to build an Oranda site into a separate directory, e.g.

```
cd ~/projects/my_cool_thing_dot_dev/
oranda build --project-root="../my_cool_thing/`
```

I'm interested in implementing this feature! I dug into the code starting with where `Build.run(...)` calls `Site::build_single(&config, None).write(Some(&config))`.

But if I understand the code correctly (and I might not!), it's best not to change the behavior of `--project-root` to match what the end-user would expect. But implementing a new flag (say, `oranda build --source`) would mean digging deep into and changing some Axo-specific library code, which I wouldn't expect Axo to accept!

It looks like the intended way-of-doing-things is to build the site alongside the repo and put it in `.gitignore`, or to point `oranda.json` to a GitHub project URL.

[...]

PR #685, opened 2024-02-05
fix: Remove project-root and config-path from documentation

lynnpepin:
For issue https://github.com/axodotdev/oranda/issues/683, where `--project-root` and `--config-path` is described in documentation but doesn't work.

This changes the docs, not the code!

### DIFF
--- a/docs/src/cli/build.md
+++ b/docs/src/cli/build.md
@@ -1,9 +1,4 @@
 # `oranda build`
 
-This command builds your oranda site. You can specify:
-
-- **The project root** (`--project-root`), in case you want to build from another directory
-- **The config path** (`--config-path`), if your configuration file is not `./oranda.json`
-
-You can also pass the `--json-only` flag in order for oranda to _only_ build an `artifacts.json` file that can
+This command builds your oranda site. You can specify the `--json-only` flag in order for oranda to _only_ build an `artifacts.json` file that can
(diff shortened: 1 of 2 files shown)


manyinsects:
on second thought, I did some spelunking, and it seems that `--project-root` does absolutely nothing (I think it used to be a stop-gap solution for integration into something else), while `--config-path` works the way you'd expect, with the caveat that it still looks for an `oranda-workspace.json` in the current directory

I think we can remove mention of `--project-root` completely, but while I don't think we should un-hide `--config-path` from the actual CLI (yet), we can at least mention it on the page (with the caveat I mentioned). does that sound good?

lynnpepin:
Sure thing! I pushed a new change into my branch.

Removing `--project-root` left a one-item list on `docs/src/cli/build.md`, so I moved those details into the paragraph where `--json-only` is discussed.  I couldn't decide if that or having a two-item list were better in `build.md`, e.g. I considered this as an alternative:

```md
# `oranda build`

This command builds your oranda site. You can specify:

- The flag `--json-only`  in order for oranda to _only_ build an `artifacts.json` file that can be read by other tools
  (or websites) for integration purposes.
- The config path `--config-path` if your configuration file is not `./oranda.json`, but oranda will still look for an
  `oranda-workspace.json` in the current directory.
  
  ```

manyinsects (approved):

